PR #6, opened 2026-05-10
Extract validator, filter, and markdown helpers from main_window

Osiris-DevWorks:
## Summary

Pulls three pure-Python helpers out of `MainWindow` so they can be unit-tested without a running `QApplication`, and shrinks `main_window.py` from 3900 lines to ~3580.

This is the first of several focused PRs distilling the genuinely useful refactors from #5 without taking on the rebrand, license change, migrator removal, or auto-updater removal that PR carries.

### New modules
- **`src/utils/applied_file_validator.py`** — `validate_applied_file(written_path, cache_dir, stock_keys=None)`. Independently re-parses the just-written `global.ini` against `base.ini` and reports missing / unexpected keys.
- **`src/utils/entry_filter.py`** — `filter_entry_indices(...)`. Closes over the seven filter inputs (column filters, category, status, hide-unmodified, favorites-only, prefix) and returns the visible-row index list. Includes a column-bounds guard that drops stale out-of-range filters with a one-shot warning instead of raising `IndexError` deep in the per-entry loop, plus a small perf micro-opt: per-call getter pre-resolution so the hot loop only computes the columns that are actually filtered (one tuple alloc instead of a 7-element list per of ~87k entries).
- **`src/gui/markdown_renderer.py`** — `markdown_to_html(text, text_color, base_color, link_color)`. Pure converter; the Qt caller passes in palette colours.

`MainWindow.{_validate_applied_file, _filtered_entry_indices, markdown_to_html}` become thin wrappers over those module functions. The deleted helpers (`create_anchor_id`, `_convert_markdown_inline`, `_convert_markdown_links`) had no callers outside `markdown_to_html`.

### Bonus: restores `tests/test_core.py` collection

`test_core.py:25` imports `utils.overrides_manager`, which doesn't exist on `1.3.0` — so the entire file silently failed to collect. Adds a 21-line `src/utils/overrides_manager.py` shim (re-exports `load_overrides` and `save_overrides` from the live modules) and updates `tests/pytest.ini` to put the project root on `pythonpath` (so `from src.X` imports resolve from tests as well as production).

`TestStringEntry` inside `test_core.py` is marked `@pytest.mark.skip` — its `StringEntry` constructor calls predate the dataclass adding `category` and `status` as required positional args, so the tests were stale anyway. Fixing them is a separate cleanup; this PR only restores file collection.

### Test results

- **Before**: 115 pass, 4 pre-existing failures in `test_pak_extraction.py::TestDataForgeCache` (`dataforge_cache_dir` is `str` not `Path`, unrelated bug), 1 collection error in `test_core.py`.
- **After**: 178 pass, same 4 pak_extraction failures, 15 skipped (the stale `TestStringEntry` block).
- **Net**: +63 passing tests, zero new failures.

The pak_extraction failures are out of scope for this PR — separate fix.

### Attribution

Module sources adapted from jonigirl's work in #5; the integration into `main_window.py`, the `pytest.ini` pythonpath change, and the `test_core.py` skip marker are this PR's additions.

## Test plan

- [x] `pytest tests/` — 178 pass, 4 unrelated pre-existing failures, 15 skipped
- [x] `python -c "from src.gui.main_window import MainWindow"` resolves
- [ ] Manual smoke test: launch app, open About + Help tabs (markdown render path), apply edits to game (validator path), apply column filters (entry_filter path)


Osiris-DevWorks:
Thanks — pushed `2ffcdb8` addressing the substantive concerns. Going point by point:

**Scope (installer.iss + VERSION.TXT)** — those were local-only commits on `1.3.0` that hadn't been pushed to origin, so they were appearing in the PR diff against the *remote* base. Pushed `origin/1.3.0` to catch up; the PR diff is now the two extraction commits only. No squash needed.

**`entry_filter._NUM_FILTER_COLUMNS = 7`** — fixed. Now imports `NUM_COLUMNS` from `src.gui.string_table_model`. Added `test_getter_count_matches_num_columns` that runs one filter per column to catch drift between `NUM_COLUMNS` and the local getter tuple if a future column is added in only one place.

**Lambda-getter perf comment** — dropped. The "amortised over ~87k entries" claim was speculative; comment now just says what it does, not why it's allegedly faster.

**Dead-code branch** — refactored. The two-step "filter then maybe-reassign then maybe-build-fns" logic collapsed into a single pass through `column_filters` that accumulates `valid_col_filters` and `bad_indices` in one go. `active_filter_fns` is now built unconditionally (empty when there are no valid filters). Per-entry loop uses `continue` instead of the `show` flag.

**`overrides_manager` shim** — removed. CLAUDE.md rule trumps the bonus fix. `tests/test_core.py:25` now imports the real functions directly:
```python
from parser.ini_parser import load_overrides
from utils.user_ini_manager import save_user_ini_dict as save_overrides
```
Two `TestOverridesManagement` tests needed `tmpdir` coerced to `Path` (the shim used to wrap `str → Path` for `save_user_ini_dict`). Tests still green.

**`pythonpath = . src`** — kept, with note. It's load-bearing: the new modules use `from src.X` internal imports (matching the rest of `src/`), and without `.` on the path the test imports of those modules can't resolve `from src.parser.ini_parser`. You're right that the codebase has two test-import styles right now (`from utils.X` for older tests, `from src.X` for the new ones), but standardising is its own refactor — happy to do it as a follow-up PR if you want.

**`AppSettings.get_cache_dir()` always called in the wrapper** — left as-is per your "cheap papercut" call.

**Installer behavior change consequences** — out of this PR's scope after the rebase. They belong in the discussion on `b19233b` directly.

**`test_entry_filter.py:103` 11-element list** — added a docstring explaining why it's meaningful (single OOB index instead of ten, exercises the validator's "drop and warn" path with one warning, not ten).

Test delta: **179 pass, 4 pre-existing pak_extraction failures unchanged, 15 skipped (the stale `TestStringEntry` block).**

PR #26, opened 2026-05-11
Implement DataForge XML caching and improve performance

Ishikudeska:
### Summary
Reduces enhancement generation time by skipping unchanged categories between game patches, and refactors the generator pipeline for clarity and maintainability.

---

#### `src/utils/dataforge_diff.py` (new file)
Adds a diff-cache utility that tracks which DataForge XML files changed between extractions and maps them to enhancement categories. Instead of regenerating all 6+ enhancement INIs on every run, only the categories whose source XMLs actually changed are re-run.

- Stores a manifest of `{ relative_path: { mtime, sha256 } }` as `.diff_manifest.json` inside the DataForge cache directory
- Uses mtime as a fast pre-filter; only computes SHA-256 when mtime differs, keeping the hot path (nothing changed) fast
- Manifests are automatically scoped per channel (LIVE / PTU / TECH-PREVIEW) since each channel has its own DataForge cache directory
- `update_manifest(cache_dir)` — call after successful extraction to snapshot the current state
- `dirty_categories(cache_dir)` — returns `None` (no manifest, run everything), `set()` (nothing changed, skip all), or `{"ships", "missions", ...}` (only re-run these)

#### `src/utils/pak_extractor.py`
Calls `update_manifest()` after a successful DataForge extraction so the next run can diff against it.

#### `src/gui/workers.py`
Consults `dirty_categories()` before launching the enhancement generators. Skips the entire generation step when nothing changed. Falls back to full regeneration if enhancement INI files are missing on disk (guards against the manifest being written before enhancements ever ran successfully).

#### `scripts/generate_enhancements_ini.py`
Refactors the generator wave from inline closures to module-level functions (`_run_gen_components`, `_run_gen_missions`, etc.) that receive shared read-only state via a context dict. No behavioral change — still runs on a `ThreadPoolExecutor` — but the functions are now independently testable, easier to read, and no longer capture `main()`'s local scope.

---

### Performance Impact

| Scenario | Before | After |
|---|---|---|
| First run / full extraction | 5–10 min | 3-4 min |
| Re-run after patch (some files changed) | 5–10 min | Proportional to changed categories |
| Re-run with no DataForge changes | 5–10 min | ~0s (skipped entirely) |

The most common case for active players — re-running after a hotfix that only touches a handful of XMLs — goes from a full regeneration to regenerating only the affected categories.

---

### Testing
- Run extraction once to populate the manifest
- Run again without re-extracting — verify log shows `Diff-cache: no DataForge changes detected, skipping enhancement generation`
- Delete one enhancement INI, run again — verify it regenerates despite the clean manifest
- Apply a game patch and re-extract — verify only the affected categories regenerate

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,6 +4,7 @@
 PyQt6>=6.10.0
 PyInstaller>=6.3.0
 pyperclip>=1.8.2
+lxml>=5.0
 
 # Testing
 pytest>=7.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 PyQt6>=6.10.0
+lxml>=5.0
 pyinstaller>=6.3.0
 pyperclip>=1.8.2
--- a/scripts/build/build_exe.py
+++ b/scripts/build/build_exe.py
@@ -153,9 +153,7 @@ common_args = [
     # is the robust fix.
     '--hidden-import=src.utils.progress_sink',
     '--hidden-import=src.utils.dataforge_patcher',
-    '--hidden-import=xml',
-    '--hidden-import=xml.etree',
-    '--hidden-import=xml.etree.ElementTree',
+    '--collect-all=lxml',
     # scripts/generate_enhancements_ini.py is loaded dynamically via
     # importlib at runtime (see EnhancementsGeneratorWorker), so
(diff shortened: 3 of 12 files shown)


Osiris-DevWorks:
this was merged into the 1.3.1 dev branch so will close this PR.

PR #48, opened 2026-05-22
generator: discover items from XML when base.ini has no entry

denis-coach:
Reverse the enhancement generator's skip logic: instead of silently discarding DataForge XML entities whose loc key is absent from base.ini, synthesize a description from XML attributes and emit them into the enhancement INI files. These entries flow through the existing "enhancements" source injection and get status "New" in the table.

Changes:
- Add _synthesize_description() helper that builds rich descriptions from XML (career/role/crew/length for ships, item type for components, tracking type for missiles, filename stem as fallback)
- scan_entity_dir() and scan_spaceships() no longer skip missing keys; they synthesize descriptions and emit discovered items
- Mission title/desc assembly synthesizes from contract debug names when loc keys are missing from base.ini
- _component_name_tag() falls back to XML ItemComponentParams for size/type when the description text lacks Size:/Grade:/Class: lines
- Add shield/cooler/powerplant/qdrive/radar to _ITEM_TYPE_ABBREV so the tagger's fallback path works for all component types
- Name tag generation synthesizes name values from XML file stems for discovered items whose name key is not in loc
- append_enhancements() guards against None existing_value

Covers all category scanners: ships, components, weapons, missiles, FPS weapons, and missions. Crafting blueprints deferred (different discovery pattern).

17 new tests in test_discovered_items.py. Full suite: 578 passed, 4 pre-existing failures in test_pak_extraction.py.

### DIFF
--- a/scripts/generate_enhancements_ini.py
+++ b/scripts/generate_enhancements_ini.py
@@ -320,6 +320,8 @@ MISSION_SEPARATOR = "\\n\\n<EM3>MISSION DETAILS</EM3>\\n"
 
 def append_enhancements(existing_value: str, enhancements_block: str,
                         separator: str = ENHANCEMENT_SEPARATOR) -> str:
+    if existing_value is None:
+        existing_value = ""
     if not enhancements_block:
         return existing_value
     # Strip any existing stats/mission details block. BP/ITEMS/BLUEPRINT DATA
@@ -448,6 +450,61 @@ def _loc_name_key(root: ET.Element) -> str | None:
     return None
 
 
+def _synthesize_description(root: ET.Element, xml_file: Path, key: str) -> str:
+    """Build a synthetic description from XML attributes when base.ini has no entry.
+
+    Returns the richest available description so discovered items still appear
+    in the table with useful context.
+    """
+    parts: list[str] = []
+
+    # 1. Item name: prefer the XML Name loc ref, fall back to filename
+    name = _loc_name_key(root)
+    if name:
+        parts.append(name)
+    else:
+        stem = xml_file.stem.replace("_", " ").strip()
+        if stem:
+            parts.append(stem)
+
+    # 2. Ship-specific attributes from VehicleComponentParams
+    vpc = _find(root, "VehicleComponentParams")
+    if vpc is not None:
+        attrs: list[str] = []
+        career = vpc.get("vehicleCareer", "")
+        if career.startswith("@"):
+            attrs.append(f"Class: {career.lstrip('@')}")
+        role = vpc.get("vehicleRole", "")
+        if role.startswith("@"):
+            attrs.append(f"Role: {role.lstrip('@')}")
+        crew = vpc.get("crewSize", "")
+        if crew:
+            attrs.append(f"Crew: {crew}")
+        bbox = _find(root, "maxBoundingBoxSize")
+        if bbox is not None:
+            length = bbox.get("y", "")
+            if length:
+                attrs.append(f"Length: {length}m")
+        if attrs:
+            parts.append(" | ".join(attrs))
+
+    # 3. Component attributes (ItemComponentParams)
+    icp = _find(root, "ItemComponentParams")
+    if icp is not None:
+        item_type = icp.get("itemType", "")
+        if item_type:
+            parts.append(f"Item Type: {item_type}")
+
+    # 4. Missile tracking signal type
+    tp = _find(root, "targetingParams")
+    if tp is not None:
+        sig = tp.get("trackingSignalType", "")
+        if sig:
+            parts.append(f"Tracking: {sig}")
+
+    return "\n".join(parts) if parts else key.replace("_", " ")
+
+
 # ── Tag emitters ─────────────────────────────────────────────────────────────
 # Each emitter pulls structured data off the description text and/or XML and
 # hands it to render_tag(). Format strings, ordering, separators, and the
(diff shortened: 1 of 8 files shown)


src/parser/ini_parser.py
```diff
@@ -190,9 +190,13 @@ def load_source_files(
             # User has an override for this key
             status = 'Modified'
         else:
-            # No user override — use source-origin-based status
             source = source_origin.get(key, base_source)
-            status = _determine_status_from_source(source, base_source)
+            # Discovered from DataForge XML — key only exists in enhancements,
+            # not in the global (base.ini) source.
+            if source == 'enhancements' and key not in base_sources.get(base_source, {}):
```
Osiris-DevWorks: `New` is being set in two places now — line 188 (`if key not in base_merged`) and this new branch.

`src/parser/CLAUDE.md` names `_determine_status_from_source` as the single source of truth for status mapping but it never actually returns `'New'` since the signature can't tell whether the key is in stock base — the call site has to know.

Could we consolidate? Something like `_determine_status_from_source(source_name, base_source, *, key_in_base: bool)` with both `New` branches moved inside. Line 188 already split it before this PR so this isn't new debt, just a natural moment to clean up if you want.

Osiris-DevWorks (commented):
Walking through this PR. Leaving one note on the parser change. More may follow as I review the rest.

src/gui/config_tab.py
```diff
@@ -290,6 +292,15 @@ def setup_ui(self):
         tools_desc.setWordWrap(True)
         tools_layout.addWidget(tools_desc)
 
+        self.include_new_cb = QCheckBox("Include new lines")
```
Osiris-DevWorks: "Include new lines" is ambiguous. Let's use something a bit more descriptive like:

- "Include discovered items"
- "Include 'New' status entries"
- "Apply XML-discovered items"

scripts/generate_enhancements_ini.py
```diff
@@ -559,31 +624,83 @@ def _component_name_tag(desc_value: str, root: ET.Element | None = None,
...
+
+    # When Class: is missing from text, try to derive from XML ItemComponentParams
+    xml_item_type = None
+    if not class_name and root is not None:
+        icp = _find(root, "ItemComponentParams")
+        if icp is not None:
+            raw_type = icp.get("itemType", "")
+            # Map CamelCase XML itemType to the space-separated form
+            # used in _ITEM_TYPE_ABBREV (e.g. "ShieldGenerator" → "Shield Generator")
+            _ITEM_TYPE_XML_MAP = {
```
Osiris-DevWorks: This literal map is rebuilt on every call to `_component_name_tag` and duplicates the CamelCase-splitting logic that `attach_class_name` already does:

```python
attach_class_name = re.sub(r"([a-z])([A-Z])", r"\1 \2", subtype)
```

Same approach would work here:

```python
if raw_type:
    xml_item_type = re.sub(r"([a-z])([A-Z])", r"\1 \2", raw_type)
```

That removes the literal map, deduplicates the CamelCase split, and makes the path generic — any new XML `itemType` would work without an entry being added.

If the explicit map is intentional as a controlled allowlist (only let these five types through the fallback), worth a comment saying so. Otherwise the regex is cleaner. Minor.

tests/test_discovered_items.py
```diff
@@ -0,0 +1,296 @@
...
+        self._make_component_xml(tmp_path, "shield2", "item_DescSHLD_B")
+        loc = {"item_DescSHLD_B": "Existing shield."}
+
+        def dummy_enhancement_fn(root):
+            return "stats"
+
+        gen_module.scan_entity_dir(
+            tmp_path, dummy_enhancement_fn, loc=loc, capture_all=False,
+        )
+        # No direct way to check the counter (it's local), but the function
```
Osiris-DevWorks: This test has no assertion — the comment says "no direct way to check the counter (it's local)" and then just runs the function. As-is it passes regardless of whether discovery actually works.

Two ways out:
1. Delete the test (it doesn't verify anything the other tests don't already cover).
2. Rewrite to assert on the observable output:

```python
result = gen_module.scan_entity_dir(
    tmp_path, dummy_enhancement_fn, loc=loc, capture_all=False,
)
assert "item_DescSHLD_A" in result  # discovered
assert "item_DescSHLD_B" in result  # in loc
```

That checks the function emits both even when the local counter itself is unreachable. Major.

tests/test_discovered_items.py
```diff
@@ -0,0 +1,296 @@
...
+        )
+        assert "vehicle_DescAEGS_AI" not in result
+
+
+# ── Parser integration ──────────────────────────────────────────────────────
+
+class TestDiscoveredItemsInParser:
+    """Covers that discovered items get 'New' status through the parser."""
+
+    def test_enhancement_only_key_gets_new_status(self, tmp_path):
```
Osiris-DevWorks, line 254: This test imports `load_source_files` but never calls it. The assertions only check that the two INIs were parsed — they don't exercise the new branch in `ini_parser.py:196` (the `if source == 'enhancements' and key not in base_sources...: status = 'New'` path), which is the actual change this PR makes to the parser.

To verify the New-status classification, something like:

```python
entries = load_source_files(
    sources_dict={"global": base, "enhancements": enhancement},
    hierarchy=["global", "enhancements"],
    user_overrides=None,
)
discovered_b = next(e for e in entries if e.key == "item_DescSHLD_B")
assert discovered_b.status == "New"

existing_a = next(e for e in entries if e.key == "item_DescSHLD_A")
assert existing_a.status == "Enhanced"  # in base AND enhancements → Enhanced, not New
```

As-is, the parser change has no test coverage exercising the new branch. Major.

Osiris-DevWorks (dismissed):

Osiris-DevWorks (approved):

Osiris-DevWorks:
@h0useRus thanks for the contribution!

PR #105, opened 2026-05-29
2.0

Ishikudeska:


scubamount:
## Comprehensive Analysis of PR #105

I did a deep audit of this PR from multiple angles (architecture, i18n design, translation coverage, edge cases, code quality). Here are the verified findings - all confirmed against the actual source code.

---

### Critical / High

**1. Dual crash handlers - both override `sys.excepthook`**
- `crash_logger.py` installs at module level (`main.py:49-50`)
- `crash_handler.py` installs inside `main()` (`main.py:70-71`)
- Both override `sys.excepthook`, creating a chain: Python → crash_handler → crash_logger → default
- Each writes crash files to **different directories** (`LOCALAPPDATA` vs `user_data_dir/logs`)
- Consider consolidating into a single handler or making crash_handler delegate to crash_logger

**2. QComboBox signals not blocked during retranslation**
- `main_window.py:2800` - `status_combo.setItemText()` fires `currentIndexChanged` during retranslation
- `config_tab.py:436-439` - `theme_combo.setItemText()` same issue
- Each `setItemText()` triggers connected signal handlers unnecessarily during language switch
- Fix: wrap with `combo.blockSignals(True)` / `try` / `finally: blockSignals(False)`

---

### Medium

**3. French `tabs.legal` key missing**
- `languages/french/ui.json` lines 9-15 - no `"legal"` key in the `tabs` section
- `languages/english/ui.json` line 15 has `"legal": "Legal"`
- The About/Legal tab will show the bare key `"tabs.legal"` instead of translated text
- Fix: add `"legal": "Juridique"` (or appropriate French term) to the `tabs` section

**4. ~48 hardcoded tooltips not wrapped in `tr()`**
- Found across 6 files: `main_window.py` (22), `config_tab.py` (10), `enhancements_tab.py` (10), `log_tab.py` (3), `tag_mapping_dialog.py` (1), `import_dialog.py` (2)
- All tooltip strings are English-only and will never update on language switch
- These are lower priority than visible UI text but worth addressing for completeness

**5. Language switch during active merge - no worker cancellation**
- `_on_language_changed()` (`main_window.py:2827-2835`) calls `perform_merge_and_reload()` without cancelling in-flight merge workers
- If a merge is already running, calling it again could cause concurrent file writes or stale results
- Consider cancelling active workers before triggering a new merge

**6. Language switch doesn't invalidate `default_values` cache**
- When the language changes and the overlay is reloaded, the `default_values` cache (used for status classification) isn't cleared
- This can cause incorrect `Modified`/`Unmodified` status until the next full reload

**7. i18n race condition on `_strings` global**
- `set_language()` modifies `_strings` and `_current_lang` without locking
- `tr()` called from a worker thread could see a partially-merged state
- Worst case: seeing English text (the fallback) - low practical impact but architecturally incorrect

**8. `_deep_merge` replaces dict with scalar silently**
- If the base has a dict at a key but the overlay has a scalar, the dict is replaced without warning
- Unlikely with well-structured JSON files but could cause hard-to-trace issues if translation files are malformed

---

### Low

**9. `print()` instead of `logger.warning()` in `ini_parser.py:57`**
- `print(f"Error parsing INI file {path}: {e}")` - should use the logging framework

**10. `SC_LANGUAGE_IDS` has orphaned `german` entry**
- Mapped to `"german_(germany)"` at `settings.py:16` but no `languages/german/` folder exists
- Dead code - harmless but confusing

**11. Portuguese BR translation typo**
- `portuguese_br/ui.json` line 89: `"Data.p4k não encontrado at: {path}"`
- English `"at:"` was left untranslated - should be `"em:"`

**12. Untranslated context menu actions - "Underline" / "Highlight"**
- `workers.py:422,425` and `main_window.py:2378,2381` - hardcoded English strings, not wrapped in `tr()`

**13. Untranslated QMessageBox dialogs**
- `"File Not Found"`, `"Empty File"`, `"Copy Filtered"`, `"Copy Error"` - all hardcoded in `main_window.py` (lines 1922, 1928, 3792, 3799, 3801)

**14. Hardcoded `"All"` in category combo**
- `main_window.py:3662` - `addItem("All")` not wrapped in `tr()`

**15. Log tab export dialog hardcoded**
- `log_tab.py:203-205` - `"Export log"` title and file filter strings not wrapped in `tr()`

**16. Language overlay silently skipped when file is empty**
- `ini_parser.py:325` - if the language file exists but `parse_ini_file()` returns empty, no warning is logged
- The user sees no indication their selected language wasn't applied

**17. `_prune_old_logs` silently swallows `OSError`**
- `crash_logger.py:34-37` - catches `OSError` and ignores it without logging a warning

---

### What I verified is correct

- **i18n architecture** - English loaded as base, target language deep-merged on top. Missing keys naturally fall back to English. This is sound.
- **`_deep_merge` semantics** - fresh dict loaded each `set_language()` call, no cross-call contamination
- **Tab bar retranslation** - all 6 tabs (including Legal) are updated in `retranslate_ui()`
- **Table headers** - retranslate via `model.retranslate()` + `headerDataChanged` signal
- **PyInstaller bundling** - `('languages', 'languages')` in datas tuple correctly includes translation files
- **Filter bar crash fix** - try/finally around layout signals is correct
- **Performance caching** - `_header_labels` and `_fav_bg` are properly invalidated on theme/palette change
- **crash_logger Qt dialog guard** - `QApplication.instance() is None` check prevents crash during early startup
- **test_pak_extraction.py argument order fix** - PR correctly fixes a pre-existing test bug
- **`set_filtered_indices` exception safety** - has try/finally (lines 210-215)
- **Spanish ui.json** - intentionally minimal with a `_comment` explaining the fallback design
- **`_status_label` in log tab** - updated dynamically via `_append_record()` with `tr()`

Ishikudeska:
Working on this all now. Thank you.

Ishikudeska:
spanish.json is intentional as we are waiting on a community member to do that translation

Issue #110, opened 2026-05-30
First-run crash: IndexError in string_table_model when base.ini is missing (no sources configured)


Osiris-DevWorks:
## Summary

On a fresh install, before DataForge / `base.ini` extraction has run, the strings table can be queried with zero entries and crashes with `IndexError: list index out of range`.

## Repro (portable 1.4.2, first run)

1. Fresh install, no `base.ini` cached yet.
2. App starts; `FileLoaderWorker` runs and fails because `…\LIVE\cache\base.ini` does not exist yet, ending in `ValueError: No sources configured`.
3. The table model's `flags()` / `entry_for_row()` is then called against an empty entry list.
4. Crash: `IndexError: list index out of range`.

## Traceback

```
File "src\gui\string_table_model.py", line 243, in flags
File "src\gui\string_table_model.py", line 215, in entry_for_row
IndexError: list index out of range
```

## Relevant log

```
src.parser.ini_parser - ERROR - Local file not found: …\LIVE\cache\base.ini
src.parser.ini_parser - ERROR - Failed to load source global …: Source global file not found
src.gui.workers - ERROR - Error loading files: No sources configured
ValueError: No sources configured
```
(the IndexError follows shortly after)

## Likely cause

When the loader fails with "No sources configured," the model is left with no entries, but the view still queries `entry_for_row(row)` / `flags()` for a row that no longer exists, and `entry_for_row` indexes the list directly with no bounds guard.

## Suggested fix

Bounds-guard `entry_for_row` (return `None` / a safe default when the row is out of range), and/or reset the model to zero rows when a load fails so the view stops requesting stale rows.

Found while triaging #100 (from the crash dumps attached there). Environment: Smart Citizen 1.4.2 portable, Windows, Python 3.11.

Osiris-DevWorks:
A 1.5.0 preview build is ready for testing: https://github.com/Osiris-DevWorks/smart-citizen/actions/runs/26695995529 (download the installer artifact from that run, then run `SmartCitizen-1.5.0-Setup.exe`).

Could you update and check whether this is fixed? Do a fresh first run before extracting DataForge and confirm it no longer crashes.

If it is resolved, let us know and we will close this out; if not, reply here with what you still see. Thanks!

Osiris-DevWorks:
Thanks for the report. On a clean first run with no sources configured and no base.ini cached yet, the strings table tried to index into an empty list and crashed with an IndexError before the window finished loading.

[...]

Issue #103, opened 2026-05-29
Pathing Issue Persists in 1.4.3


Osiris-DevWorks:
I uninstalled 1.4.1 via Windows uninstaller. When I installed 1.4.2, my favorites were missing. I had to go into configuration and point the user directory to where 1.4.1 had them installed. I'm not sure why Smart Citizen didn't persist the location from the previous install during the installation process. We need to make sure it's saving these paths in the Windows Registry so that they persist between installs, uninstalls, etc... 

If a user changes their user directory manually, that should be persisted to the windows registry, and the contents of the previous user directory need to be migrated over.

Osiris-DevWorks:
A 1.5.0 preview build is ready for testing: https://github.com/Osiris-DevWorks/smart-citizen/actions/runs/26695995529 (download the installer artifact from that run, then run `SmartCitizen-1.5.0-Setup.exe`).

Could you update and check whether this is fixed? Change your Smart Citizen data folder on the Config tab and confirm you are prompted to copy your existing data (favourites, backups) into the new folder.

If it is resolved, let us know and we will close this out; if not, reply here with what you still see. Thanks!

Osiris-DevWorks:
When I ran this, I got the message that the uninstaller was not present. This appears to be a regression. I suspect it happens when you allow the installer to uninstall first. Needs further investigation.

Osiris-DevWorks:
This worked as intended but could use improvement as a user experience: 

Recommend having a confirmation dialog before any actions are taken. The dialog should explain to the user if they accept they are going to have the option to move/copy existing files to new location.

Osiris-DevWorks:
fixed in 2.0

PR #116, opened 2026-05-30
WIP: cargo mission info — root-cause + implementation plan (#102)

Osiris-DevWorks:
## Heads up: this PR is a plan, not a completed fix

The other five 1.5.0 PRs in this batch (#110, #101, #98, #103, #97) are complete, tested fixes. **This one is deliberately a draft investigation + implementation plan.** #102 changes the mission-rewards output for hundreds of missions, and shipping *wrong* XP/rep would be worse than the missing block the issue reports, so it needs a fixture-validated implementation rather than a rushed change late in a long batch.

## What's in it

`docs/dev/cargo-mission-info-plan.md` — the full root-cause analysis and the fix plan:

- Cargo/delivery missions live in `missionbroker/pu_missions` (not `contractgenerator`), so the contractgenerator-based enrichment never covers them.
- They carry inline `missionResultReputationRewards` whose `reward` UUIDs are **not** landing in `reputation_lookup`, so `_extract_mission_xp` returns 0 and the augmentation pass skips them as `no_rep_data` (~645 missions).
- The referenced reward definitions **do exist** under `reputation/rewards/missionrewards_reputation` (e.g. `36d5786c… → reputationAmount=500`), but only 54 load into the lookup.
- Fix direction: complete `reputation_lookup` coverage (and any one-hop indirection), so the existing augmentation emits the info block for cargo missions with **no change to the output format**.
- Validation plan using the existing `diff_mission_rewards_channels.py` / `kraken` / `missions_*.csv` ground-truth tooling, plus a unit test.

## Why not just implement it now

The remaining open question (exactly why those reward UUIDs miss the lookup: coverage vs. one-hop indirection) needs a real-data sweep of the LIVE cache and validation against the ground-truth fixtures. I'd rather hand you a precise, correct plan than a generator change I couldn't fully validate. Happy to implement from this plan as the next focused task.

Addresses #102.

### DIFF
--- a/docs/dev/cargo-mission-info-plan.md
+++ b/docs/dev/cargo-mission-info-plan.md
@@ -1,0 +1,89 @@
+# Plan: cargo missions missing mission information (#102)
+
+Status: **investigation + plan** (not yet implemented). Unlike the other
+1.5.0 fix PRs, this one captures the root-cause analysis and the intended fix
+so the implementation can be done and validated against real data with care.
+Producing *wrong* mission XP/rep is worse than the current missing-info state,
+so this deserves a deliberate, fixture-backed implementation rather than a
+rushed change.
+
+## Symptom
+
+Most Cargo / delivery missions show no mission-information block (rep tier, XP,
+etc.) in their description. Reporter (Narull) saw details on ~3 of 39 cargo
+missions. Confirmed still present on 1.4.2.
+
+## What we already know
+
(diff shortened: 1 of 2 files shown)


Osiris-DevWorks:
Closing this without merging. The implementation plan now lives on the tracking issue #102 (posted there as a comment) rather than as a code-less doc in the release branch, so #102 is the single source for the cargo-mission-info work. #102 stays open as the tracker for a future minor; reopen/convert this when the build starts.

PR #118, opened 2026-05-30
2.0

Osiris-DevWorks:
Replacement for #105, which GitHub closed automatically when the `2.0` branch was renamed to `release/2.0.0`. Same branch, same commits (`d15d2f5`), just the new name, no change in intent. Original PR opened by @Ishikudeska.

This is the long-lived 2.0 integration target merging into `main`.

---

## Testing checklist (changes since v1.4.2)

Covers only what changed since the last released version (v1.4.2). Internal test-only fixes, docs, and merge commits are left out on purpose. Test against the installer and portable artifacts built from the `build-installer` / `build-portable` labels on this PR.

### New features
- [x] **Commodity Tagging (#97)** — Enhancements tab: enable the Commodities category and Generate. Crafting and collection items get their tags; an item that is both crafting and collection shows the combined form. Change the tag form options and confirm they take effect on the next Generate.
- [x] **Data folder migration (#103)** — Config tab: change the Smart Citizen data folder to a new path. The existing user.ini, cache, backups, and enhancement INIs move to the new location and the app keeps working.
- [x] **Per-field mission details (#121)** — Enhancements tab, "Mission detail fields": uncheck individual fields (Mission Type, Difficulty, Spawns, Reputation, Blueprints) and Generate. Each unchecked field drops from mission bodies. The separate "Blueprint Tag" checkbox removes the [BP] / [BP?] marker from mission titles independently of the body section.
- [x] **Toolbar More menu (#128)** — The toolbar shows Apply, Editor, Help, Tutorial, and a More dropdown. More holds Restore Backup, Clear Localization / Clear Cache, Import / Export, and Open Localization Dir. Confirm every action still works from the menu.
- [x] **Language switching (#30)** — Config tab Language selector: switch to French, then Brazilian Portuguese. The UI (tabs, toolbar, Config and Enhancements labels) retranslates live with no restart. Untranslated keys fall back to English (both are partial translations, so some English remains). Switch back to English and confirm everything reverts.

### Fixes
- [x] **First-run crash (#110)** — Launch on a clean install with no sources configured and no cached base.ini. The app opens to an empty table instead of crashing.
- [x] **Blueprint-list type tag (#101)** — Enable the component Type element in the Tag Builder, Generate, and check a mission blueprint list: component entries carry the type tag (Shield, Cooler, Power Plant, Quantum Drive, Radar), matching standalone component names.
[...]

PR #119, opened 2026-05-31
fix: validate SC install paths to reject stale registry values

scubamount:
## Problem

Stale registry values like `SmartCitizen 1.4.1` are returned as the SC install root, causing the app to show wrong default paths in the Config tab.

**Root cause**: `get_sc_install_root()` and `get_game_install_path()` return raw registry values without checking they actually contain Star Citizen channel subdirectories (LIVE, PTU, EPTU, HOTFIX, TECH-PREVIEW).

### src/utils/settings.py
- Added `_is_valid_sc_root(path)` helper that validates a path contains at least one channel subdirectory
- Fixed `get_sc_install_root()`: validate saved `SC_INSTALL_ROOT` and legacy `GAME_INSTALL_PATH` before returning them
- Fixed `get_game_install_path()`: validate legacy `GAME_INSTALL_PATH` and registry `sc_directory` before returning them

### installer.iss
- Added `IsValidSCPath(Path)` function that accepts both channel paths (ending in LIVE/PTU/etc.) and root paths (containing channel subdirectories)
- Wrapped all 5 registry reads in `InitializeWizard()` with validation guards

### tests/test_sc_install_root.py
- Updated 2 existing tests to mock `_is_valid_sc_root` (they test cross-check logic, not validation)
- Added 7 new unit tests for `_is_valid_sc_root()` edge cases (nonexistent paths, stale values, valid roots, invalid strings)

## Behavior Change

| Scenario | Before | After |
|----------|--------|-------|
| Registry has `SmartCitizen 1.4.1` | Returned as install root | Rejected, falls through to auto-detect |
| Registry has valid root (e.g. `StarCitizen` with LIVE inside) | Returned | Returned (unchanged) |
| Registry has channel path (ends in LIVE) | Returned | Returned (unchanged) |

## Verification

Fix is correct, minimal, and achieves the goal without breaking valid installations.

All callers handle empty string returns gracefully - no regressions for valid paths.

CI compatibility: tests updated, no new dependencies, PyInstaller build unaffected.

Merge compatibility: verified clean merge into PR #118 (2.0 branch). The 2.0 branch modifies `settings.py` in non-overlapping regions (i18n/language support only). Both PRs insert code at module level after the logger line but as independent additions (different symbols), so Git auto-merges cleanly. No changes needed in `installer.iss` or `test_sc_install_root.py` on the 2.0 branch.

### DIFF
--- a/installer.iss
+++ b/installer.iss
@@ -617,24 +617,23 @@ end;
 
 function IsValidSCPath(const Path: String): Boolean;
 var
-  ExtPath: String;
+  BasePath: String;
+  LastName: String;
 begin
   { Returns True if Path looks like a valid Star Citizen install path.
     Accepts either:
       - a root containing channel subdirectories (LIVE, PTU, etc.), or
       - a channel path (last component is a channel name like LIVE).
     Guards against stale registry values like 'SmartCitizen 1.4.1'. }
-  ExtPath := Path + '\';
-  if (LowerCase(ExtractFileName(RemoveBackslash(Path))) = 'live') or
-     (LowerCase(ExtractFileName(RemoveBackslash(Path))) = 'ptu') or
-     (LowerCase(ExtractFileName(RemoveBackslash(Path))) = 'eptu') or
-     (LowerCase(ExtractFileName(RemoveBackslash(Path))) = 'hotfix') or
-     (LowerCase(ExtractFileName(RemoveBackslash(Path))) = 'tech-preview') then
+  BasePath := RemoveBackslash(Path) + '\';
+  LastName := LowerCase(ExtractFileName(BasePath));
+  if (LastName = 'live') or (LastName = 'ptu') or (LastName = 'eptu') or
+     (LastName = 'hotfix') or (LastName = 'tech-preview') then
     Result := True
   else
-    Result := DirExists(ExtPath + 'LIVE') or DirExists(ExtPath + 'PTU') or
-              DirExists(ExtPath + 'EPTU') or DirExists(ExtPath + 'HOTFIX') or
-              DirExists(ExtPath + 'TECH-PREVIEW');
+    Result := DirExists(BasePath + 'LIVE') or DirExists(BasePath + 'PTU') or
+              DirExists(BasePath + 'EPTU') or DirExists(BasePath + 'HOTFIX') or
+              DirExists(BasePath + 'TECH-PREVIEW');
 end;
 
 procedure InitializeWizard();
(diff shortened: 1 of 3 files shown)


installer.iss
```diff
@@ -652,22 +674,39 @@ begin
...
+    end;
   end;
 
   { Fall back to previously saved sc_directory / game_install_path in the
-    NEW node, then the LEGACY node. }
+    NEW node, then the LEGACY node.  Each value is validated to ensure it
+    actually looks like a valid SC install path (either ends with a channel
+    name, or contains channel subdirectories). }
   if DefaultPath = '' then
   begin
```
Osiris-DevWorks, line 686: This nesting has a catch. The outer `if` fires whenever `sc_directory` exists and isn't empty, even when it's the stale value we want to reject. `IsValidSCPath` then returns false, `DefaultPath` stays empty, and because the outer branch already matched, the entire `else if` chain below is skipped, including the `C:\Program Files\...\StarCitizen\LIVE` auto-detect. The net effect: when `sc_directory` is stale, the page comes up blank instead of falling back to a sensible default. And since the installer writes `sc_directory` and `game_install_path` to the same value (lines 424-425), this is likely to hit the exact users this PR is meant to help.

Cleaner to fold the validity check into each condition so an invalid value simply doesn't match and the chain keeps going:

```pascal
if RegQueryStringValue(HKCU, NewRegPath, 'sc_directory', SavedPath) and (SavedPath <> '') and IsValidSCPath(SavedPath) then
  DefaultPath := SavedPath
else if RegQueryStringValue(HKCU, NewRegPath, 'game_install_path', SavedPath) and (SavedPath <> '') and IsValidSCPath(SavedPath) then
  DefaultPath := SavedPath
else if ...
```

That restores the fallthrough and lets you drop the nested `begin/end` blocks entirely.

installer.iss
```diff
@@ -615,6 +615,28 @@ begin
...
 
+function IsValidSCPath(const Path: String): Boolean;
+var
+  ExtPath: String;
+begin
+  { Returns True if Path looks like a valid Star Citizen install path.
+    Accepts either:
+      - a root containing channel subdirectories (LIVE, PTU, etc.), or
+      - a channel path (last component is a channel name like LIVE).
+    Guards against stale registry values like 'SmartCitizen 1.4.1'. }
```
Osiris-DevWorks, line 627: Small one: if `Path` already ends in a backslash, this gives you `...\StarCitizen\\LIVE`. `DirExists` shrugs that off on Windows, but it's inconsistent with the channel-name branch just above, which runs through `RemoveBackslash`. Normalize once: `ExtPath := RemoveBackslash(Path) + '\';`. And while you're in here, `LowerCase(ExtractFileName(RemoveBackslash(Path)))` is recomputed five times, so you could pull it into a local for readability.

src/utils/settings.py
```diff
@@ -9,6 +9,24 @@
...
+        p = Path(path)
+        if not p.is_dir():
+            return False
+        return any((p / ch).is_dir() for ch in (
+            "LIVE", "PTU", "EPTU", "HOTFIX", "TECH-PREVIEW"
+        ))
+    except (OSError, ValueError):
+        return False
+
 class AppSettings:
```
Osiris-DevWorks, line 45: `black` and `flake8` want two blank lines before a top-level class, so this needs one more blank line above it. CI didn't catch it because no checks ran on the PR, but the formatter will.

src/utils/settings.py
```diff
@@ -529,7 +547,14 @@ def get_game_install_path() -> str:
         # Legacy path stored under the old key.
         saved = AppSettings.settings().value(AppSettings.GAME_INSTALL_PATH, "")
         if saved:
-            return saved
+            saved_path = Path(saved)
+            if saved_path.name.upper() in (c.upper() for c in AppSettings.AVAILABLE_CHANNELS):
```
Osiris-DevWorks: This channel-name check now lives in three places: here, the `sc_directory` branch below, and the existing one in `get_sc_install_root`. That's the point where it's worth pulling out a small `_path_ends_in_channel(path)` helper next to `_is_valid_sc_root` so they stay in step.

Separate, worth a line in the PR description rather than a code change: because `_is_valid_sc_root` checks the channel folder is on disk right now, a valid path on an unplugged external drive (or a correct root where the channel folder doesn't exist yet) gets dropped and re-auto-detected. Probably fine for us, I just want it to be a deliberate choice rather than a surprise later.

tests/test_sc_install_root.py
```diff
@@ -122,3 +136,58 @@ def _fake_exists(self):
 
         result = AppSettings.get_sc_install_root()
         assert result == r"D:\Games\StarCitizen"
+
+
+class TestIsValidScRoot:
```
Osiris-DevWorks, line 141: These cover `_is_valid_sc_root` on its own really well. The one gap: both edited tests up top mock it to `True`, so nothing runs a stale value through `get_sc_install_root` and checks it actually gets dropped. That end-to-end path is the whole point of the fix and the thing a future refactor could quietly break. Could you add one: set `SC_INSTALL_ROOT` to a `tmp_path` folder with no channel subdir, block the Program Files fallbacks, and assert the getter doesn't hand it back?

Minor while you're here: tagging the new stale-value test with `pytest.mark.regression` matches how we mark these (see `test_favorite_prefix_whitespace.py`), and a `tmp_path` dir would be a tidier stand-in than `C:\Windows`.

Osiris-DevWorks (changes requested):
Thanks for digging into this, @scubamount. The diagnosis is dead on: stale values like `SmartCitizen 1.4.1` were getting trusted as install paths, and guarding the read side is the right call. The app-side getters and the Star Citizen directory page are handled cleanly here.

A few things need a second pass before it can land. One is a real bug in the installer that bites in the exact case this PR is fixing, so I want to flag it up front rather than bury it.

**1. The installer's fallback chain stops falling back.** When `sc_directory` holds the stale value, the outer `if` still matches, the validity check fails, and the whole `else if` chain below (including the Program Files auto-detect) gets skipped. So the user ends up with a blank Star Citizen directory field in precisely the scenario we're trying to fix. Details and a suggested rewrite are in the inline comment.

**2. The fix isn't tested where it matters.** The two edited tests mock `_is_valid_sc_root` to `True`, so nothing actually runs a stale value through the getter and checks it gets dropped. That end-to-end path is the point of the change. One test covers it.

**3. Smaller things:** a missing blank line that `black`/`flake8` will want, a channel-name check that's now duplicated three times, a double-backslash edge in the new installer helper, and a regression marker on the new test. All in the inline comments.

**One scope note, not on you:** the reported screenshots are the Data Location and DataForge Cache pages, which read `user_data_dir` and `cache_dir`. This PR doesn't touch those, so it's a partial fix. I'll open a separate issue for those two pages and the root cause of how the stale value got written in the first place.

None of this is heavy lifting and the core idea is solid. Once the installer chain and the test are sorted, I think this is good to go.

scubamount:
Thanks for the thorough review. All 5 inline items are addressed in 2 new commits on top of `d8424b1` (now 3 commits ahead of `release/1.5.0`):

**1. Installer fallback chain** — fixed in `3f468ee`. Folded `IsValidSCPath` into each `RegQueryStringValue` condition so a stale value simply does not match and the chain falls through to the Program Files auto-detect. The `sc_install_root` branch (line ~675) had the same nested bug and is fixed the same way. Also normalised the trailing-backslash edge in the helper (`RemoveBackslash(Path) + '\'`) and pulled `LowerCase(ExtractFileName(...))` into a local so it's no longer recomputed 5 times.

**2. End-to-end stale-value test** — added in `d753dfb` as two `@pytest.mark.regression` tests. One drives `SC_INSTALL_ROOT`, the other drives `GAME_INSTALL_PATH`. Both set the value to a `tmp_path/SmartCitizen 1.4.1` directory, block the Program Files auto-detect, and assert `get_sc_install_root()` returns `""`. The existing `TestInstallRootCrossCheck` tests still mock `_is_valid_sc_root` to `True` so they continue to exercise the cross-check logic, not the gate.

**3. Smaller items** — also in `d753dfb`:
- Added the missing second blank line above `class AppSettings:`
- Extracted `_path_ends_in_channel(path)` helper next to `_is_valid_sc_root`. Replaced the inline check in all **4** call sites (`get_game_install_path` ×2, `get_sc_install_root` cross-check, `get_sc_install_root` legacy fallback).

Both commits are now on `scubamount:fix/sc-install-path-validation` and visible on this PR (3 commits ahead of `release/1.5.0` total). PR is ready for re-review. Happy to squash the two review-response commits into the original if you'd prefer a single commit on the PR.

Osiris-DevWorks (approved):

Issue #30, opened 2026-05-13
Adding support for multiple languages


Ishikudeska:
Adding support for multiple languages, starting with:
- Spanish
- Russian
- Chinese

Goal 1:
Finding reputable community made translations -
Looking on spectrum and reddit for highly used and recommended language .ini's

Goal 2:
Backend Implementation

Goal 3:
UI Work

More features can/will be added, but we need to see a reason to do so.

PinkFloyd1213:
Even if it's not at the top of the priority list right now, here are two very reliable and reputable sources for the French translation of Star Citizen. Both are well-known in the community and consistently kept up to date:

* **Scefra (Speed0u):** [https://github.com/SPEED0U/Scefra](https://github.com/SPEED0U/Scefra)
* **Circuspes:** [https://traduction.circuspes.fr/download/](https://traduction.circuspes.fr/download/)

These projects are the "gold standard" for French community translations and would be excellent references for high-quality `.ini` files.

Ishikudeska:
> Even if it's not at the top of the priority list right now, here are two very reliable and reputable sources for the French translation of Star Citizen. Both are well-known in the community and consistently kept up to date:
> 
> * **Scefra (Speed0u):** https://github.com/SPEED0U/Scefra
> * **Circuspes:** https://traduction.circuspes.fr/download/
> 
> These projects are the "gold standard" for French community translations and would be excellent references for high-quality `.ini` files.

We personally would like to work directly with those making these translations. This is so we can avoid using AI to translate things.

PinkFloyd1213:
You're absolutely right, and I can definitely reassure you on that front before any contact is made: both projects I mentioned (Scefra and Circuspes) are 100% hand-translated by volunteers.

The philosophy behind these two projects has always been to prioritize "hand-crafted" quality over AI-generated content. They focus on maintaining the nuances and specific terminology of the Star Citizen universe, which is exactly why they are so highly regarded by the French community.

Ishikudeska:
I personally think that AI has its uses in translation (common phrases and words, basically any "bulk work" that could just be automated with a script) but nuanced things like fictional names need to be human done

Osiris-DevWorks:
> Even if it's not at the top of the priority list right now, here are two very reliable and reputable sources for the French translation of Star Citizen. Both are well-known in the community and consistently kept up to date:
> 
> * **Scefra (Speed0u):** https://github.com/SPEED0U/Scefra
[...]

PR #143, opened 2026-06-03
fix: bundle languages/ so the frozen UI shows text, not raw keys

Osiris-DevWorks:
## The bug

The installed 2.0.0 build showed raw i18n keys everywhere (`branding.title`, `toolbar.apply_btn`, ...) instead of translated text.

## Cause

`scripts/build/build_exe.py` (the build path the release checklist uses to produce the installer) never added the `languages/` directory to its PyInstaller `--add-data` list. So the frozen app shipped with **no** `ui.json`. At runtime `get_resource_path("languages/<lang>/ui.json")` missed, `i18n._load_file` returned `{}`, and every `tr()` call fell back to the bare key, rendering the whole UI as dot-keys.

The i18n feature (#30) added `('languages', 'languages')` to `SmartCitizen.spec`, but `build_exe.py` carries its own separate `--add-data` list and was never updated. The two bundling lists drifted. This is a frozen-only failure, so it doesn't show in `python src/main.py` (dev reads `languages/` straight from the source tree).

## Fix

Add the matching `--add-data` entry to `build_exe.py`. The two lists (spec and build script) are now identical again: VERSION, docs×3, assets, patches, **languages**, scripts.

## Verification

Built the onedir with `build_exe.py` and confirmed the files now land in the bundle:

```
dist/SmartCitizen/_internal/languages/english/ui.json
dist/SmartCitizen/_internal/languages/french/ui.json
dist/SmartCitizen/_internal/languages/portuguese_br/ui.json
dist/SmartCitizen/_internal/languages/spanish/ui.json
```

These resolve through `get_resource_path` (`_MEIPASS/languages/...`) the same way the already-working `patches/` and `assets/` bundles do, so `set_language()` loads `ui.json` and `tr()` resolves keys.

## Note

This is a release blocker: every frozen 2.0.0 build before this is unusable (raw keys). It should land before any further tester installer artifacts are cut from this branch.

### DIFF
--- a/scripts/build/build_exe.py
+++ b/scripts/build/build_exe.py
@@ -132,6 +132,7 @@ about_file   = os.path.join(root_dir, 'docs', 'ABOUT.md')
 help_file    = os.path.join(root_dir, 'docs', 'HELP.md')
 legal_file   = os.path.join(root_dir, 'docs', 'LEGAL.md')
 patches_dir  = os.path.join(root_dir, 'patches')
+languages_dir = os.path.join(root_dir, 'languages')
 enhancements_script = os.path.join(root_dir, 'scripts', 'generate_enhancements_ini.py')
 
 common_args = [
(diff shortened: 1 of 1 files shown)


Osiris-DevWorks:
Folding this directly into release/2.0.0 instead. For this session we're committing fixes straight to the release branch and pushing them as one batch, so the #118 tester preview rebuilds only once. Same commit (c2f9225), no change to the fix.

PR #146, opened 2026-06-04
fix: Add no space option for Commodities

denis-coach:


Osiris-DevWorks:
Appreciate the thought, but I do not agree that "no space" should be an option in this case. I do not see a scenario where "CFCollection" is preferable to having a separator.

PR #150, opened 2026-06-06
fix: keep newly chosen SC install root when switching install locations

hkstrongside:
## What

`set_sc_install_root()` could silently discard a freshly chosen install
root and revert to the previous one, breaking setup for anyone whose Star
Citizen install isn't at the C: default.

## Symptom

Browse to the correct SC install root (e.g. on D:). The path shows in the
Config-tab field, but the Channel dropdown doesn't populate, the warning
"LIVE isn't installed under this root — pick another channel" appears, and
P4K extraction reads the old (C:) location.

## Root cause

`set_sc_install_root()` refreshed the legacy `GAME_INSTALL_PATH` by reading
back through `get_channel_install_path()` → `get_sc_install_root()`. That
getter runs a cross-check that trusts `GAME_INSTALL_PATH` over a disagreeing
`SC_INSTALL_ROOT`. At that instant `GAME_INSTALL_PATH` still holds the *old*
value, so the cross-check treats the new root as the mistake and reverts
`SC_INSTALL_ROOT` back to the stale location (persisting the revert).

## Fix

Compose the active-channel path from the value passed into the setter
instead of reading it back through the cross-checking getter. Both keys are
written to the new location, so the cross-check never sees a transient
disagreement.

## Tests

Adds `TestSetInstallRootSwitchesLocation` to `tests/test_sc_install_root.py`:
C:→D: switch persists the new root, the synced legacy path respects the
active channel, and the empty path clears the legacy key. The two switch
tests fail on `main` and pass with the fix. Full suite: 742 passed, 16 skipped.

## Note on target branch

The branching model targets the active `release/X.Y.Z` branch, but none is
open post-2.0.0, so this targets `main`. Happy to retarget to a release
branch when one opens — it's a patch-scope bug fix.

### DIFF
--- a/src/utils/settings.py
+++ b/src/utils/settings.py
@@ -1786,9 +1786,16 @@ class AppSettings:
         consistent without manual bookkeeping at each call site.
         """
         AppSettings.settings().setValue(AppSettings.SC_INSTALL_ROOT, path)
-        AppSettings.set_game_install_path(
-            AppSettings.get_channel_install_path() if path else ""
+        # Compose the active-channel path from the value we were just handed,
+        # NOT via get_channel_install_path()/get_sc_install_root(). That getter
+        # runs a cross-check against the still-stale GAME_INSTALL_PATH and, when
+        # the user switches install locations (e.g. C: default → D:), reverts
+        # SC_INSTALL_ROOT back to the old root before we get a chance to refresh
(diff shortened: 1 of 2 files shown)


Osiris-DevWorks (approved):
@hkstrongside thanks for the contribution, looks good!

PR #168, opened 2026-06-18
Add ship sort-order column for ASOP arrangement (#142)

### LINKED ISSUE #142: ASOP Ship Arrange
*Synced from Discord: [ASOP Ship Arrange](https://discord.com/channels/1438175448420057323/1511685236528779424)*

suggestion: easy menu to "number" ships (modify name?) to put them in specific order in ASOP. ?

Osiris-DevWorks:
## What this does

Closes #142. Adds an easy way to **number ships so they line up in a chosen order in ASOP**.

Each ship row now has a sort-order number (`#` column) next to the favorite star. You pick a number, and Smart Citizen tucks it into the ship's name so it sorts that way in the game's ASOP terminal. Mark a few favorites, number them 1, 2, 3, and they'll show in that order.

## How it works

- The favorite star and the sort number both live in the ship's `custom_value`. The favorite prefix comes first, then a **zero-padded two-digit number** (ex: `*01Avenger`). Zero-padding matters: ASOP sorts text, so without it `10` would come before `2`. With it, `01 < 02 < ... < 10` holds.
- New `#` column with a 0-99 spin-box editor. **0 clears** the number (shown blank). Editable on Ship rows only; other categories show nothing there.
- You can sort the table by the `#` column too: numbered ships first (ascending), the rest after.

## Implementation notes

- The prefix math lives in a new Qt-free `src/utils/ship_sort_prefix.py` (`get_order` / `set_order`), so it's unit-tested without a `QApplication`.
- Wiring: `COL_ORDER` in the table model, `OrderSpinBoxDelegate` in `workers.py`, header/filter/column-width setup in `main_window.py`, the `entry_filter` getter tuple, and a `col_order` key in each active `ui.json`.

## Tests

- New `tests/test_ship_sort_prefix.py` (18 cases) covers get/set, favorite interaction, clear-to-empty collapse, and padding.
- `tests/test_entry_filter.py` updated for the shifted column indices, plus a new order-column getter test.
- Full suite: **764 passed, 16 skipped**.

## Manual verification still recommended

GUI behavior (double-click to edit the `#` cell, ASOP ordering in-game) isn't covered by automated tests, per the repo's manual-GUI-testing convention. Worth a quick smoke test before merge: number two favorite ships and confirm the order in ASOP.


Osiris-DevWorks:
Ready for review. Two fixes landed on this branch after the initial column work, both found via local verification:

- **Delegate lifetime** (`6da18b9`): parent the Custom Value and Sort Order delegates to the table. `setItemDelegateForColumn` doesn't take ownership and PyQt keeps only one reference per method, so the second (unparented) delegate was letting the Custom Value delegate get garbage-collected, silently dropping select-all and the EM3/EM4 menu on that column.
- **Numeric ship names** (`a2e0a43`): ships whose name starts with two digits (300i, 600i, 890 Jump, 85X, ...) were read as carrying a sort order, which put a phantom number in the # column and chopped the leading digits off the name on edit (`*300i` -> `*0i`). Order tokens now use a `NN-` separator (`05-Avenger`) so a bare numeric name is never mistaken for an order. Regression tests cover the real offender names.

Suite: 788 passed, 16 skipped.

Issue #158, opened 2026-06-14
Add [ACE] tag for ace-pilot missions


Osiris-DevWorks:
## Request
Flag missions that spawn an ace pilot with an add-on title tag, coexisting with the existing `[BP]` and rep/XP tags, plus a MISSION DETAILS line.

## Investigation (scunpacked-data)
- Ace pilots appear as a spawn group named **`AcePilotShip`** — in 157 contracts.
- Every `AcePilotShip` group is `Min:1, Max:1, Weight:1` → **the ace always spawns when the group is present.** There is **no probability field** for the spawn. (The only `Chance` field in the contract data belongs to blueprint reward pools.)
- **There is no "% chance" to display** — what players perceive as a chance is which contract variant the mission board rolls (board RNG), not anything in the localization data.
- Ace variants are almost always **their own contract with their own loc key** (difficulty baked into the key, e.g. `_S_`, `_VH_`, `_H_`, `_VE_`): **52 of 59** DescriptionKeys are exclusive to ace contracts.
- **7 of 59** keys are shared between ace and non-ace contracts (one displayed string reused across location variants where only some spawn an ace). Example: `HeadHunters_ShipAmbush_VH_desc_001` covers Nyx (ace), Pyro (no ace), Stanton (no ace).

## Design (decided)
- **Presence-based, no percentage.**
- Title add-on tag: **`[ACE]`** for keys exclusive to ace contracts; **`[ACE?]`** for the shared/ambiguous keys (mirrors the `[BP?]` convention). Appended alongside `[BP]`/rep/XP tags.
- MISSION DETAILS line: **"Ace Pilot: Yes"**, gated by a per-field show/hide setting like other mission-detail fields.
- Detection reuses the generator's existing `classify_spawn_group` "Ace Pilots" flag (`scripts/generate_enhancements_ini.py`).

Part of the release/2.1.0 blueprint-feature set (siblings: BP filters, Owned tag).

Osiris-DevWorks:
Shipped in v2.1.0: missions that spawn an ace pilot carry an [ACE] title tag, [ACE?] when only some variants do.

PR #192, opened 2026-06-28
fix(missions): component tag fallback, full show_fields gating, ace detail (#158 #160 #163)

Ishikudeska:
## Summary

Four issues addressed across two commits.

---

### #160 -- `_component_name_tag` fallback no longer emits grade-only tags

The fallback path (used when the `Class:` lookup fails) previously emitted
a tag like `[S2-A]` whenever a grade was known, even when no Item Type
abbreviation could be resolved. The class prefix is the only part that
tells a player what kind of item they are looking at; a grade alone is
meaningless as a tag. Changed the guard from
`if not (type_abbrev or grade_letter)` to `if not type_abbrev` and
simplified the parts-list construction. Items with a recognised grade but
an unrecognised type now return `None` and fall through to the raw name.

**Covered by:** `test_tagger_fallback_grade_only_no_known_item_type` in
`test_blueprint_pools.py` -- expectation updated from `[S2-A]` to `None`.

---

### #163 -- `enhancements_mission` now respects all mission-detail field toggles

`enhancements_mission` is the code path for standalone `pu_missions` and
`entities/missions` XMLs (not the contractgenerator loop). It previously
only gated the spawns bucket via a bare `show_spawns: bool` parameter
added in an earlier partial fix, leaving `mission_type`, `difficulty`, and
`reputation` always emitted regardless of user preference.

Replaced `show_spawns: bool = True` with `show_fields: dict | None = None`
and added a local `_show()` helper that defaults missing keys to `True` (so
an unconfigured call gets the full body, exactly as before). All four
fields are now gated. Both call sites in `_run_gen_missions` updated from
`show_spawns=_show_spawns` to `show_fields=_mdf`, making the `pu_missions`
and contractgenerator paths use the same settings dict end-to-end.

**Covered by:** three `TestSpawnGatingInEnhancementsMission` tests in
`test_spawn_classifier.py` updated from `show_spawns=...` to
`show_fields={"spawns": ...}`; all `test_mission_detail_fields.py` tests.

---

### #158 -- "Ace Pilot: Yes" detail line in mission description bodies

The `[ACE]`/`[ACE?]` title-level tags were already in HEAD. This adds the
companion line in the MISSION DETAILS body: emits `Ace Pilot: Yes` when
the aggregated spawn breakdown for that description contains a non-zero
`Ace Pilots` hostile group, gated by `_show("ace")` so it respects the
user's per-field toggle independently of the title tag.

---

### New feature -- "Standardize earnable ship names" (PYX / WIK)

CIG's loc keys for exec-hangar (PYX) and Wikelo (WIK) ship variants share
the same `vehicle_Name*` prefix as their pledge-store equivalents, so
Smart Citizen has no way to distinguish them in the ship descriptions
table. A player who owns both the standard F8C Lightning and the
exec-hangar PYX version sees two entries with the same display name.

Adds an opt-in toggle ("Standardize earnable ship names") on the
Enhancements tab. When enabled, a static lookup table of ~35 entries is
applied at write time just before `ships_desc_enhancements.ini` is
written, appending a `PYX` or `WIK` suffix (plus a variant tag for WIK
ships: War / Work / Stealth) to every earnable-only variant. Entries with
an empty-string value are placeholder stubs and are skipped. Feature is
off by default -- existing users see no change until they opt in.

Ships covered:

**PYX exec-hangar (6):** F7A Hornet Mk II, Cutlass Black, Meteor
Collector, F8C Lightning, Corsair, Syulen

**WIK Wikelo (29):** Fortune, Guardian QI, Pulse, Ursa Medivac, Nox,
C1 Spirit, Polaris -- plus War/Work/Stealth variants including Sabre
Firebird, Idris-P, F8C Lightning, Ares Ion/Inferno, A2 Hercules,
L-21/L-22 Wolf, Starlancer TAC/MAX, Guardian/MX, Constellation Taurus,
Apollo Triage, Meteor, Scorpius, RAFT, Intrepid, Golem, Prowler Utility,
Prospector, Zeus CL/ES

**Files changed:**

- `scripts/generate_enhancements_ini.py`: `EARNABLE_SHIP_NAME_OVERRIDES`
  dict + `standardize_earnable_ship_names` param on `main()`
- `src/utils/settings.py`: `STANDARDIZE_EARNABLE_SHIP_NAMES` key +
  `get/set_standardize_earnable_ship_names()`
- `src/gui/enhancements_tab.py`: `_standardize_ship_names_check` checkbox
  with tooltip, wired to settings
- `src/gui/workers.py`: param plumbed through `EnhancementsGeneratorWorker`
- `src/gui/main_window.py`: setting read and passed to worker

## Test plan

- [x] `pytest tests/test_blueprint_pools.py` -- `test_tagger_fallback_grade_only_no_known_item_type` returns `None`
- [x] `pytest tests/test_spawn_classifier.py` -- three `TestSpawnGatingInEnhancementsMission` tests pass with new `show_fields` API
- [x] `pytest tests/test_mission_detail_fields.py` -- all field-gate tests pass
- [x] `pytest tests/` -- full suite (937 passed, 16 skipped)
- [ ] Manual: enable "Standardize earnable ship names", run Generate Enhancements, confirm PYX/WIK entries appear in the ships table with the correct suffixes
- [ ] Manual: disable the toggle, regenerate, confirm names revert to CIG originals

### DIFF
--- a/scripts/generate_enhancements_ini.py
+++ b/scripts/generate_enhancements_ini.py
@@ -786,13 +786,10 @@ def _component_name_tag(desc_value: str, root: ET.Element | None = None,
     if not type_abbrev and class_name:
         type_abbrev = _ITEM_TYPE_ABBREV.get(class_name)
 
-    if not (type_abbrev or grade_letter):
+    if not type_abbrev:
         return None
 
-    parts: list[str] = []
-    if type_abbrev:
-        parts.append(type_abbrev)
-    parts.append(f"S{size}")
+    parts: list[str] = [type_abbrev, f"S{size}"]
     if grade_letter:
         parts.append(grade_letter)
     return f"[{'-'.join(parts)}]"
@@ -2592,7 +2589,7 @@ def _extract_mission_flags(root: ET.Element) -> list[str]:
 
 def enhancements_mission(root: ET.Element, reputation_lookup: dict[str, int] | None = None,
                          rep_xp_label: str = _DEFAULT_REP_XP_LABEL,
-                         show_spawns: bool = True,
+                         show_fields: "dict | None" = None,
                          spawn_ambiguous_keys: "set[str] | None" = None) -> str:
     """Extract mission/contract reward stats (aUEC + Reputation XP) and flags.
 
@@ -2605,25 +2602,29 @@ def enhancements_mission(root: ET.Element, reputation_lookup: dict[str, int] | N
     lines = []
     reputation_lookup = reputation_lookup or {}
 
+    def _show(f: str) -> bool:
+        return bool((show_fields or {}).get(f, True))
+
     try:
         loc_key = _mission_loc_key(root) or _loc_key(root)
         lines.append(f"<EM4>Engagement Type:</EM4> {_classify_mission_engagement(loc_key)}")
 
         flags = _extract_mission_flags(root)
-        lines.append(f"<EM4>Mission Type:</EM4> {', '.join(flags) if flags else 'Standard'}")
+        if _show("mission_type"):
+            lines.append(f"<EM4>Mission Type:</EM4> {', '.join(flags) if flags else 'Standard'}")
 
         difficulty = _extract_difficulty(root)
-        if difficulty:
+        if difficulty and _show("difficulty"):
             lines.append(f"<EM4>Difficulty (1-7):</EM4> {difficulty}")
 
         total_rep_xp = _extract_mission_xp(root, reputation_lookup)
-        if total_rep_xp > 0:
+        if total_rep_xp > 0 and _show("reputation"):
             lines.append(f"<EM4>{rep_xp_label}:</EM4> +{total_rep_xp:,}")
 
         # Extract spawn/wave counts — bucketed Hostiles / Friendlies /
         # Objectives / Unknown rather than the pre-1.4.1 single-tally
         # Enemies + Non-hostiles. Empty buckets are dropped by the formatter.
-        # #163: gated by the Hostiles mission-detail toggle. The contractgen
+        # #163: gated by per-field show_fields (Hostiles toggle). The contractgen
         # path gates this too (via _show("spawns")); pu_missions / entities
         # missions reach the table through here, so without the gate salvage
         # contracts kept showing hostiles after the user turned them off.
@@ -2633,7 +2634,7 @@ def enhancements_mission(root: ET.Element, reputation_lookup: dict[str, int] | N
         # — a single body can't honestly show one count for both. The
         # consistent buckets (e.g. Friendlies "Salvageable Ships") still show.
         _ambiguous = loc_key in spawn_ambiguous_keys if spawn_ambiguous_keys else False
-        if show_spawns:
+        if _show("spawns"):
             _bd = _extract_spawn_counts(root)
             if _ambiguous:
                 _bd[SPAWN_HOSTILE] = {}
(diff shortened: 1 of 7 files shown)


scripts/generate_enhancements_ini.py
```diff
@@ -5949,6 +6002,10 @@ def _builder() -> dict[str, str]:
     logger.info("Writing output files…")
     _flush()
     if _want("ship_descs"):
+        if standardize_earnable_ship_names:
+            applied = {k: v for k, v in EARNABLE_SHIP_NAME_OVERRIDES.items() if v}
+            out_ships.update(applied)
+            logger.info(f"Earnable ship name overrides: applied {len(applied)} entries")
```
Osiris-DevWorks, line 6008: Nice work on this, and the `if v` placeholder skip is a clean touch. Two things before it can go in.

Can we get a small unit test around this block? Just enough to lock the two behaviors: empty-string entries get skipped, and the non-empty ones actually land in `out_ships`. It's simple logic, but it's new logic, and our standard is that new paths ship with a test.

The other piece is the override table itself. Those ~37 loc keys are hand-typed, and a unit test can't catch a misspelled key (a bad key just silently does nothing). So this one needs a real in-game pass too: flip the toggle on, generate, and confirm the PYX/WIK names show up in the ships table and in-game before we call it done.

scripts/generate_enhancements_ini.py
```diff
@@ -5344,6 +5394,8 @@ def _show(_field: str) -> bool:
                 details_lines.append(f"<EM4>Difficulty (1-7):</EM4> {all_difficulties[0]}")
             if _show("spawns"):
                 details_lines.extend(_format_spawn_lines(agg_spawns))
+            if _show("ace") and bool(agg_spawns.get(SPAWN_HOSTILE, {}).get("Ace Pilots", 0)):
+                details_lines.append("<EM4>Ace Pilot:</EM4> Yes")
```
Osiris-DevWorks, line 5398: This adds the line on the contractgen path, where the spawn aggregation lives, so that tracks. One thing to double-check: do any ace missions come through the `enhancements_mission` path (pu_missions / entities) instead? If they do, they'd get the `[ACE]` title tag but not this body line, which is the same split we just fixed for #163. If ace missions are always contractgen, ignore this. Either way, a tiny test that the line shows when Ace Pilots > 0 and the toggle is on would be good.

src/gui/enhancements_tab.py
```diff
@@ -256,6 +256,21 @@ def _build_enhancements_group(self) -> QGroupBox:
         )
         gl.addWidget(self._stats_prepend_check)
 
+        self._standardize_ship_names_check = QCheckBox("Standardize earnable ship names")
```
Osiris-DevWorks, line 259: Small one, and no need to fix in this PR: the label and tooltip here are plain English while the buttons below use `tr()`. It matches the checkboxes right above it, so it's not new drift, but let's add a `tr()` key for it at the docs-sync pass so it's ready for translation.

scripts/generate_enhancements_ini.py
```diff
@@ -2605,25 +2602,29 @@ def enhancements_mission(root: ET.Element, reputation_lookup: dict[str, int] | N
     lines = []
     reputation_lookup = reputation_lookup or {}
 
+    def _show(f: str) -> bool:
+        return bool((show_fields or {}).get(f, True))
```
Osiris-DevWorks, line 2606: Just noting, not a change request: this `_show` is a second copy of the one in `_run_gen_missions`. They close over different dicts so they can't easily share, and two copies sits inside our DRY rule. Leaving the note here so the next person knows it's intentional.

Osiris-DevWorks (changes requested):
Thanks for batching these, Ishi. The three bug fixes (#158, #160, #163) look right to me, and the #160/#163 test updates lock them in. I've left a few inline notes.

Requesting changes mostly to get a test on the new feature.

Issue #221, opened 2026-07-06
Generated stats for mining lasers using invalid markups in Mobiglass.


XaileiShiv:
Looking at the vehicle loadout manager and looking at the generated stats for mining lasers the use of "<-EM4->" is present.

A square character between the optimal and maximum range of the laser in meters.  In game format not supporting whatever character was being used there.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e8a94eaf-a4fd-43cc-8ea9-fd016d75aa10" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/60acecda-7e4c-45a2-bf2e-6c58b7784cb9" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/628ad4b9-93c2-4b22-a029-e156ab18d452" />

Stealrull:
Issue fixed in new release 2.2.0, if it appears again please make a new report.

PR #231, opened 2026-07-13
Bug fixes: install path, progress dialogs, tag/blueprint matching, mining laser stats

Stealrull:
## Summary
10 independent bug fixes bundled onto one branch for easy review:

- Config tab's install-path Browse button opened Qt's default folder instead of the configured path
- Progress dialog auto-closed mid-way through multi-phase operations
- DataForge cache-snapshot progress bar looked frozen at 100% while the manifest was still writing
- Declining the Data.p4k extract prompt at startup threw error popups instead of skipping the load
- Class/Grade Blueprint Tracker facets went missing for categories using append-style tags
- WinError 3 crash on long install paths during DataForge patching
- `sync_key_variants` overwrote correctly-tagged loc values with an unenhanced sibling key (the "7SA Concord" / Taiga cooler bug)
- Config tab install-path field didn't reflect an auto-detected legacy install
- Mining laser/salvage tool in-game stats showed literal `<EM4>` tags and a box glyph (GitHub bug report)
- String Editor's filter row showed on every tab instead of just the String Editor

Each commit is self-contained and independently revertable.

## Test plan


Stealrull:
Appended 6 more bug fixes from continued testing: String Editor preview-pane move + height halving (follow-ups to the filter-row move already here), the Bengal-mounted gatling turret patch clobbering the APAR Revenant S4's stats, hostile spawn counts leaking from a handler's intro contract onto sibling missions (the "Handle Security Threat" wrong-hostiles report), WinError 3 long-path crashes across every DataForge cache consumer (deeply nested portable installs), and packing the enhancement category grid left so the tab doesn't scroll horizontally.

src/merger/ini_merger.py
```diff
@@ -154,22 +154,30 @@ def sync_key_variants(merged_dict: Dict[str, str]) -> None:
...
+            # _SCItem-suffixed keys are always the "internal" duplicate. That
+            # broke the Taiga cooler case: item_NameCOOL_WCPR_S02_Taiga_SCItem
+            # held the correctly tagged value while its non-_SCItem sibling
+            # item_Name_COOL_WCPR_S02_Taiga was the never-enhanced bare one —
+            # so pre-filtering to "non-_SCItem only" picked the bare one every
+            # time. Comparing length across *all* variants with no pre-filter
+            # is a no-op for the genuine same-entity variant case this
+            # function targets (the generator's own sibling mirroring already
+            # gives those matching values, so length ties and either wins).
+            preferred_key = max(variants, key=lambda v: len(merged_dict[v]))
```
Osiris-DevWorks: One interaction worries me here. sync_key_variants runs after user overrides are applied in merge_sources_by_hierarchy, and enhancement siblings carry long stat blocks. So if someone edits a variant key to anything shorter than its sibling's value, max-by-length picks the sibling and quietly reverts their edit on the next merge. User overrides are supposed to apply last and survive (root CLAUDE.md, Merge hierarchy). The old rule at least let an edit to the non-_SCItem key always win; now survival depends on character count. Suggestion: pass the user-override key set into sync_key_variants and prefer any user-edited variant, falling back to longest. If we decide variant keys are never user-edited in practice, fine, but let's make that call on purpose.

src/merger/ini_merger.py
```diff
@@ -154,22 +154,30 @@ def sync_key_variants(merged_dict: Dict[str, str]) -> None:
...
-                if not var.lower().endswith('_scitem'):
-                    preferred_key = var
-                    synced_value = merged_dict[var]
-                    break
-
-            # If all have _SCItem (unlikely), just use the first
-            if synced_value is None:
-                preferred_key = variants[0]
-                synced_value = merged_dict[preferred_key]
+            # Prefer the longest value among ALL variants — no _SCItem
```
Osiris-DevWorks, line 176: This is the second time this selection rule has been quietly wrong, and sync_key_variants still has no test anywhere in tests/. The two bug reports are ready-made fixtures: the 7SA pair (short untagged sibling must lose), the Taiga pair (the _SCItem variant holds the correct value and must win), plus a tie case to lock in the no-op behavior for genuine same-entity variants. A small test class here means the next heuristic change can't re-break these.

src/merger/ini_merger.py
```diff
@@ -154,22 +154,30 @@ def sync_key_variants(merged_dict: Dict[str, str]) -> None:
...
+            # build the canonical form means two genuinely DIFFERENT CIG loc
+            # keys can collapse onto the same canonical string even though
+            # they hold different text — e.g. item_Name_SHLD_BEHR_S01_7SA
+            # (fully tagged/enhanced) and item_NameSHLD_BEHR_S01_7sa (a
+            # distinct, never-enhanced sibling CIG ships with just the bare
+            # stock name) both canonicalize to "itemnameshldbehrs017sa".
+            # Picking an arbitrary "first" variant let the short, unenhanced
+            # sibling silently overwrite the correctly tagged one.
+            #
+            # An earlier version of this fix pre-filtered to non-_SCItem
```
Osiris-DevWorks: The first paragraph of this comment earns its keep (the canonical collapse is genuinely surprising). This second one is history of the fix, not guidance for the next reader. I'd fold the Taiga example into the first paragraph as a plain counterexample (_SCItem variants can hold the correct value, so no pre-filter) and drop the rest. Not blocking.

Osiris-DevWorks (commented):
Reviewed the sync_key_variants fix. The canonical-collapse diagnosis is solid and the length rule handles both reported pairs. Three notes inline: one interaction with user overrides I want settled before merge, one ask for tests, one comment trim.

src/utils/win_paths.py
```diff
@@ -0,0 +1,29 @@
...
+    directory plus the per-channel cache layout, the destination path can
+    exceed the legacy 260-char ``MAX_PATH``, and ``shutil``/``os``/lxml raise
+    ``WinError 3`` ("The system cannot find the path specified") even
+    though the path is otherwise valid (#221). The ``\\\\?\\`` prefix tells
+    the Win32 API to skip that check (paths up to ~32,767 chars). No-op on
+    non-Windows platforms and on paths that are already prefixed.
+    """
+    if sys.platform != "win32":
+        return str(path)
+    p = str(Path(path).resolve())
```
Osiris-DevWorks: resolve() does more than the prefix needs. The \?\ prefix just wants a fully qualified, normalized path, and os.path.abspath gives exactly that. resolve() also follows symlinks and junctions, so a cache root behind a junction (or a subst drive) quietly becomes its real path everywhere downstream: logs, error messages, anything we write back. Probably harmless here, but it's a behavior change beyond what the fix needs. Either switch to abspath or add a one-liner saying symlink resolution is deliberate.

src/utils/win_paths.py
```diff
@@ -0,0 +1,29 @@
...
+    """Return *path* as a Windows extended-length path string when needed.
+
+    DataForge's ~28k-file entity tree uses long CIG-authored filenames
+    (e.g. ``softlock_terminal_standard_lowtech_commoditykiosk_console_
+    transfers_1_straight_a.xml``); once nested under a user's install
+    directory plus the per-channel cache layout, the destination path can
+    exceed the legacy 260-char ``MAX_PATH``, and ``shutil``/``os``/lxml raise
+    ``WinError 3`` ("The system cannot find the path specified") even
+    though the path is otherwise valid (#221). The ``\\\\?\\`` prefix tells
+    the Win32 API to skip that check (paths up to ~32,767 chars). No-op on
```
Osiris-DevWorks, line 18: Tiny docstring nit: "No-op on... paths that are already prefixed" isn't quite true, since an already-prefixed path still goes through resolve() first and can come back altered. Functionally fine, just worth aligning the wording with the code once the resolve/abspath question above is settled.

Osiris-DevWorks (commented):
win_paths.py looks good: one shared helper instead of per-module fixes, correct UNC handling, and it ships with its own test file. Two small notes inline.

src/utils/blueprint_meta.py
```diff
@@ -67,11 +67,14 @@
...
-# than a fixed pattern, because the tag's separator, element order, and which
-# elements appear are all Tag-Builder-configurable (see parse_component_tag).
+# The bracketed component tag, e.g. "[MIL-S3-B]" or the user-reconfigured
+# "[CMP.S1.B.PW]". Parsed by tokenizing the contents rather than a fixed
+# pattern, because the tag's separator, element order, and which elements
+# appear are all Tag-Builder-configurable (see parse_component_tag). Matched
+# leading OR trailing since the category's placement setting can put it on
+# either side (mirrors normalize_item_name's leading/trailing handling).
 _TAG_BRACKET_RE = re.compile(r"^\s*\[([^\]]+)\]")
+_TRAILING_TAG_BRACKET_RE = re.compile(r"\[([^\]]+)\]\s*$")
```
Osiris-DevWorks, line 77: This fix restores a whole facet feature for append-configured categories, but nothing locks it in. test_blueprint_meta.py::test_parse_component_tag already covers the leading shapes and is the exact home for this. One line does it:

```python
assert parse_component_tag("Balandin [MIL-S3-B]") == ("MIL", "S3", "B")
```

Ideally with a both-sides case too, to pin the leading-wins rule. Cheap insurance.

src/utils/blueprint_meta.py
```diff
@@ -100,21 +103,26 @@ class BlueprintItem:
...
+    prepend) or trailing (``Norfield [MIL-S3-B]``, append) — a category
+    configured to append lost Class/Grade extraction entirely until this
+    matched normalize_item_name's leading/trailing symmetry. The tokens
+    inside the ``[...]`` are split on any non-alphanumeric separator and
+    classified: an ``S?<digits>`` token is the size, a lone A-F letter is the
+    grade, and the first multi-letter token is the class (type codes like
+    ``PW`` come after the class in the tag, so they don't win). Missing
+    pieces are ``None``.
 
     Limitation: a single-letter (Short-style) class code can't be told apart from
```
Osiris-DevWorks, line 121: Worth adding a line here: the class token is still unvalidated, so a component name that happens to end with a non-tag bracket (a stock name ending in "[Prototype]", say) would grow a bogus PROTOTYPE class facet now that trailing brackets match. The leading-only regex had the same weakness and the call-site gate makes it unlikely in practice, so a docstring note is all this needs.

Osiris-DevWorks (commented):
The trailing-tag fix is correct, and the call-site gate keeps the false-positive surface small. Two notes inline: one quick test to add, one docstring line.

src/utils/dataforge_diff.py
```diff
@@ -145,10 +147,25 @@ def update_manifest(
...
+    # this module hits the same limit during the post-extraction snapshot).
+    cache_dir = Path(win_long_path(cache_dir))
     snapshot = _build_snapshot(cache_dir, progress_callback=progress_callback)
+    # The hash sweep above reports 100% the instant it finishes, but writing
+    # ~28k entries out is itself not instantaneous — without this the bar
+    # sat pinned at "100%"/full width while the JSON write was still going,
+    # reading as frozen. Drop back to indeterminate with a distinct label so
+    # it's visibly still doing something until the write actually completes.
+    if progress_callback:
+        progress_callback(0, 0, "Writing diff manifest…")
```
Osiris-DevWorks, line 161: Good call dropping to indeterminate here instead of letting the bar sit pinned at 100%. One thing to confirm: this (0, 0, label) convention only works if every consumer of the callback treats total == 0 as "go indeterminate" and never divides by it. I'll check the worker/GUI side, but if anything downstream computes a percentage from total, this is a divide-by-zero waiting for a slow disk.

src/utils/dataforge_patcher.py
```diff
@@ -164,7 +166,15 @@ def _apply_single_patch(
...
     target_path = records_root / target_rel
-    if not target_path.exists():
+    # DataForge paths under a deep portable/tester install directory can
+    # exceed the 260-char MAX_PATH — lxml's ET.parse/tree.write open the
+    # file directly (unlike pathlib's own I/O, which Windows tolerates past
+    # the limit in more cases) and raise a raw WinError 3 that aborted the
+    # whole DataForge extraction. The \\?\ long-path prefix sidesteps it —
+    # see win_paths.win_long_path (originally added for pak_extractor.py's
+    # copy/cleanup step, #221; this module hits the same limit).
+    long_target = Path(win_long_path(target_path))
```
Osiris-DevWorks: This wraps per patch file inside the loop, while the other three consumers (and the updated CLAUDE.md) follow the wrap-the-root-once-at-entry pattern. Patch counts are small so the per-file resolve() cost is nothing, but wrapping records_root once in apply_patches would match the documented pattern and shrink this function. Consistency is the ask, not performance.

src/utils/dataforge_patcher.py
```diff
@@ -164,7 +166,15 @@ def _apply_single_patch(
         return
 
     target_path = records_root / target_rel
-    if not target_path.exists():
+    # DataForge paths under a deep portable/tester install directory can
```
Osiris-DevWorks: This same 7-line WinError 3 story now appears here, in dataforge_diff.py, and at the other call sites. The full explanation already lives in win_paths.win_long_path's docstring, which is the right home for it. Per the DRY rule in root CLAUDE.md, the third repeat is the signal: each call site just needs one line, something like "long-path safety, see win_paths.win_long_path".

Osiris-DevWorks (commented):
The long-path wraps in dataforge_diff.py and dataforge_patcher.py both check out. I verified manifest keys are built with relative_to(cache_dir), so the prefix can't leak into stored keys and old manifests stay valid. Notes inline.

src/utils/dataforge_diff.py
Osiris-DevWorks, line 161: Separate small note: this label is another hardcoded English string the i18n layer never sees. It matches the existing snapshot labels in this module, so no change needed here, just adding it to the pile for whenever we sweep progress strings into tr().

src/utils/settings.py
```diff
@@ -2049,6 +2050,30 @@ def get_sc_install_root() -> str:
...
 
+        # Installer-written registry key (older flow) — same source
+        # get_game_install_path() falls back to. Without this, a fresh
+        # profile whose only trace of the install is this key sees
+        # get_game_install_path() (and thus extraction) resolve correctly
+        # while the Config tab, which reads only this method, stayed blank
+        # until some other code path happened to call get_game_install_path()
+        # first and persist GAME_INSTALL_PATH as a side effect.
+        try:
+            reg_path = r'Software\Osiris DevWorks\SC Localization Editor'
```
Osiris-DevWorks: This block is now a near-twin of the one in get_game_install_path (line 1059), and both hard-code the same two literals: the legacy registry path and 'sc_directory'. Root CLAUDE.md's DRY rule calls this out directly: magic literals in 2+ places get a named constant immediately. A small helper like _read_legacy_installer_sc_directory() returning the raw value (or None) would collapse both blocks; each caller keeps its own root-vs-channel interpretation, which you got right here (stripping to the root where the older branch keeps the channel path).

src/utils/settings.py
```diff
@@ -2049,6 +2050,30 @@ def get_sc_install_root() -> str:
...
+            registry_key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, reg_path)
```
Osiris-DevWorks: If QueryValueEx raises, CloseKey never runs and the handle leaks. winreg.OpenKey supports the context manager protocol, so `with winreg.OpenKey(...) as registry_key:` fixes it in one line. The pre-existing block at 1059 has the same flaw; the shared helper suggested above erases both at once.

src/utils/settings.py
```diff
@@ -2049,6 +2050,30 @@ def get_sc_install_root() -> str:
...
+            winreg.CloseKey(registry_key)
+            if sc_directory:
+                root = None
+                if _path_ends_in_channel(sc_directory):
+                    root = str(Path(sc_directory).parent)
+                elif _is_valid_sc_root(sc_directory):
+                    root = sc_directory
+                if root:
+                    AppSettings.settings().setValue(AppSettings.SC_INSTALL_ROOT, root)
+                    return root
```
Osiris-DevWorks: Tiny one: WindowsError is just an alias of OSError in Python 3, so this tuple is one exception twice. Matches the older block's style, so I'd fold the cleanup into the same helper extraction rather than touch it separately.

Osiris-DevWorks (commented):
pak_extractor.py is the cleanest of the long-path consumers: wrap-once at entry, both copytree sides prefixed, short pointer comments. No notes there. The settings.py fallback is right, but it lands as a copy-paste of the block in get_game_install_path. Notes inline.

src/gui/main_window.py
```diff
@@ -3751,6 +3759,16 @@ def _on_startup_sync_finished(self):
         if p4k_extraction_started:
             return
 
+        # User declined the extraction prompt (or it didn't fire, e.g. unp4k
+        # missing) and there's still no cached base.ini. Loading sources now
+        # would just fail with "file not found" — skip it instead of
+        # surfacing error popups for a state the user just chose to leave.
+        if not base_ini.exists():
+            self.statusBar().showMessage(
+                "No strings loaded — extract from Data.p4k on the Config tab first"
```
Osiris-DevWorks: Two things on this string. First, it's user-facing chrome that never goes through tr(), so every non-English user gets English here; it needs a ui.json key. Second, this exact literal already exists at the 0-entries branch of the load path (line 4083), so we now have the same user-facing sentence hard-coded twice. Per the DRY rule that's the point where it gets one home: a tr() key used by both sites. And while you're in there, swap the em dash for a colon; house style keeps em dashes out of product copy.

src/gui/main_window.py
```diff
@@ -3595,8 +3599,12 @@ def _cleanup_pending_old_cache(self) -> None:
             logger.info(f"Queued cache cleanup target already absent: {old_path}")
             return
         try:
-            import shutil
-            shutil.rmtree(old_path, ignore_errors=False)
+            # _robust_rmtree (not a raw shutil.rmtree) so this survives both
+            # transient Windows locks and a deep old cache path past the
+            # 260-char MAX_PATH (long-path-wraps internally — see
+            # win_paths.win_long_path).
+            from src.utils.pak_extractor import _robust_rmtree
```
Osiris-DevWorks: This is now the second place main_window imports _robust_rmtree (the DataForge clear-cache path at line 1881 does the same). Once a leading-underscore helper is imported across module boundaries twice, it isn't private anymore. I'd promote it to robust_rmtree in pak_extractor and update both call sites. Using it here instead of a raw shutil.rmtree is the right call, though.

src/utils/CLAUDE.md
```diff
@@ -19,10 +19,11 @@ Utility layer: settings, build-mode flag, P4K extraction, enhancement support mo
...
...Is accept it past the legacy 260-char `MAX_PATH`, with no machine-wide config needed (unlike the inbox "Enable Win32 long paths" policy, which requires admin rights and isn't guaranteed on for any given user — issue #221). The DataForge XML tree's long CIG-authored filenames routinely cross this limit once nested under a deep portable/tester install path (or a zip double-extracted into an already same-named folder), and `WinError 3` ("cannot find the path specified") fires even though the file exists. No-op on non-Windows and on already-prefixed paths. Every DataForge-cache consumer wraps its own root path once at its entry point so every path it derives (via further `/` joins, `rglob`, `ET.parse`, etc.) inherits long-path safety for free — see `pak_extractor.py` (copy/cleanup + the two entry points above), `dataforge_patcher.py` (patch target reads), `dataforge_diff.py` (`update_manifest`/`dirty_categories` — where the follow-up crash was actually reported, the post-extraction snapshot/hash pass), and `scripts/generate_enhancements_ini.py`'s `main()` (wraps `forge_dir` once, covering 20+ `ET.parse` sites and 18+ `rglob` walks). `tests/test_win_long_path.py` locks all of the above.
```
Osiris-DevWorks, line 26: One claim here is ahead of the code: "Every DataForge-cache consumer wraps its own root path once at its entry point." dataforge_patcher.py wraps per patch file inside _apply_single_patch, not once at entry. If the patcher gets the root-wrap refactor from my earlier comment, this sentence becomes true as written; otherwise it should say the patcher wraps per target. Doc and code should land agreeing, whichever way you go.

tests/test_win_long_path.py
```diff
@@ -0,0 +1,183 @@
...
+same-named folder, doubling a path segment) pushes individual DataForge XML
+paths past the legacy 260-char Windows MAX_PATH — even though the file is
+right there, Windows file APIs refuse it with a raw ``WinError 3`` ("cannot
+find the path specified"). ``win_paths.win_long_path`` sidesteps this with
+the ``\\\\?\\`` extended-length prefix, which Windows honors on every
+supported version with no machine-wide configuration needed (unlike the
+inbox "Enable Win32 long paths" policy, which requires admin rights and
+isn't guaranteed to be on for any given user).
+
+This module already covered ``pak_extractor.py``'s copy/cleanup step and
```
Osiris-DevWorks: Small wording trip: "This module already covered..." reads like it means this file, which is brand new. I think you mean the win_paths helper (or the original #221 fix) already covered pak_extractor and the patcher. Worth a one-word fix so the next reader doesn't go hunting for tests that were never here.

Osiris-DevWorks (changes requested):
Finished the full pass. There's a lot of good work in here: the progress-dialog autoClose fix is the right diagnosis, win_paths.py is a clean shared helper with real tests, the Bengal patch description is a model of how to document a data fix, and rewriting the dispatch-routing test as an equality proof is genuinely better than the tag-presence check it replaced.

Requesting changes on four things before this lands:

1. The sync_key_variants longest-value rule can clobber user overrides (it runs after they're applied). That needs either a guard or a deliberate decision, plus tests. (Comment on ini_merger.py.)
2. sync_key_variants and the trailing-tag fix in blueprint_meta.py both change behavior with zero new tests. Both are one-liners to lock in.
3. The sc_directory registry block is now copy-pasted with magic literals in two places. Extract the helper. (Comments on settings.py.)
4. The "No strings loaded" status string is untranslated and now duplicated. (Comment on main_window.py.)

Everything else inline is take-it-or-leave-it polish. Nice haul of fixes overall.

src/merger/ini_merger.py
Stealrull: Added the guard as suggested: `sync_key_variants` now takes `user_edited_keys`, and `merge_sources_by_hierarchy` passes the `user_overrides` keys through automatically. An edited variant always wins over the length heuristic; when both variants of a pair were user-edited, falls back to longest-of-those (no ordering signal between two deliberate edits). Also added `tests/test_ini_merger.py` — this function had zero coverage before; it now locks the 7SA and Taiga cases plus the new guard (bc471c8 on this branch).

src/merger/ini_merger.py
Stealrull: Added `tests/test_ini_merger.py`: `TestSyncKeyVariantsRealDataShapes` covers the 7SA pair (short untagged sibling must lose), the Taiga pair (`_SCItem` variant holds the correct value and must win), the tie case, and an unrelated-keys no-op; `TestSyncKeyVariantsUserEditGuard` covers the user-edit guard added for the other comment on this PR. In commit bc471c8.

src/merger/ini_merger.py
Stealrull: Trimmed in cf14f60 — folded the Taiga example into the first paragraph as a counterexample and dropped the history-of-the-fix paragraph.

src/utils/win_paths.py
Stealrull: Switched to `os.path.abspath` in cf14f60 — docstring now explains why (no symlink/junction resolution, unlike `resolve()`).

src/utils/win_paths.py
Stealrull: Reworded in cf14f60 now that abspath is used: "already prefixed... checked before any transformation, so this really is a no-op for those".

src/utils/blueprint_meta.py
Stealrull: Added in cf14f60 — `test_parse_component_tag_trailing_placement` and `test_parse_component_tag_leading_wins_when_both_present` in tests/test_blueprint_meta.py.

src/utils/blueprint_meta.py
Stealrull: Added the docstring note in cf14f60 documenting the unvalidated-class-token limitation.

src/utils/dataforge_diff.py
Stealrull: Checked the consumer side: `AnimatedProgressDialog.set_progress` (src/gui/workers.py) branches on `total <= 0` first and switches to indeterminate (`setRange(0, 0)`) before ever touching `total` in a division — the `setValue(min(completed, total))` line only runs in the `else` branch where `total > 0`. No divide-by-zero path.

src/utils/dataforge_patcher.py
Stealrull: Moved to wrap `records_root` once in `apply_patches()` in cf14f60, matching the other consumers.

src/utils/dataforge_patcher.py
Stealrull: Trimmed both call sites to one-line pointers at `win_paths.win_long_path`'s docstring in cf14f60 (DRY third-repeat rule).

src/utils/CLAUDE.md
Stealrull: dataforge_patcher.py got the wrap-once refactor (see the comment above), so the CLAUDE.md claim is accurate as written now — no doc change needed.

tests/test_win_long_path.py
Stealrull: Fixed in cf14f60 — now points at the win_paths helper (#221) instead of reading self-referentially.

src/utils/settings.py
```diff
@@ -2049,6 +2063,24 @@ def get_sc_install_root() -> str:
...
                 return legacy
 
+        # Installer-written registry key (older flow) — same source
+        # get_game_install_path() falls back to. Without this, a fresh
+        # profile whose only trace of the install is this key sees
+        # get_game_install_path() (and thus extraction) resolve correctly
+        # while the Config tab, which reads only this method, stayed blank
+        # until some other code path happened to call get_game_install_path()
+        # first and persist GAME_INSTALL_PATH as a side effect.
+        sc_directory = AppSettings._read_legacy_installer_sc_directory()
```
Osiris-DevWorks, line 2073: This fallback reads the live registry with nothing between it and the test suite. On a machine that still has the legacy `SC Localization Editor` tree (mine does: `sc_directory` points at `...\StarCitizen\LIVE`), three tests in `test_sc_install_root.py` fail:

- `test_neither_set_falls_through`
- `TestIsValidScRoot::test_stale_sc_install_root_dropped_by_getter`
- `TestIsValidScRoot::test_stale_game_install_path_dropped_by_getter`

All three expect `""` and block auto-detection by mocking `Path.exists` on the RSI paths. That mock can't stop this step: `_path_ends_in_channel` is a pure string check, so a registry value ending in `\LIVE` returns its parent without ever touching the filesystem.

The behavior itself is right (it mirrors `get_game_install_path()`, and the Config-tab blank-field bug was real). The tests just need to isolate it:

1. Monkeypatch `AppSettings._read_legacy_installer_sc_directory` to return `None` in that file's fixture. The helper you extracted makes that a one-liner.
2. Add one test covering the new step with the helper mocked to a channel-suffixed value, so the resolution order stays locked.

Worth knowing why you did not see it: a fresh profile has no legacy tree, so the suite runs clean there. It only goes red on a machine that ever ran the pre-0.9 installer, which is exactly the population this fallback exists for.

Osiris-DevWorks (changes requested):
Re-reviewed the four follow-up commits. Every item from the first pass is addressed, and addressed well:

- **sync_key_variants** now threads the user-edited key set through, so a user's edit always beats the longest-value heuristic. The tie-break for two edited variants is documented, and the function finally has test coverage (the 7SA and Taiga shapes plus the new guard).
- **win_long_path** switched to os.path.abspath with a clear docstring on why resolve() was wrong, and dataforge_patcher now wraps records_root once at entry.
- **settings.py** collapsed the twin registry reads into `_read_legacy_installer_sc_directory()`, and the context manager fixes a handle leak the old code had on the exception path. Nice catch.
- The **"No strings loaded"** message is a `tr()` key at both call sites now.

The xml_path_index v2 bump is a good self-catch too. That stale-pickle crash would have hit every cache built before this branch.

One new problem before this merges: the legacy-registry fallback added to `get_sc_install_root()` breaks three tests in `test_sc_install_root.py` on any machine that still has the old SC Localization Editor registry tree. Mine does, and the suite goes red there (3 failed, 1087 passed). Details inline. Fix the test isolation and this is ready.

src/utils/settings.py
Stealrull: Fixed in 82b48eb — the shared `json_backend` fixture now also stubs `_read_legacy_installer_sc_directory` to `None`, so the whole file is isolated from whatever's actually in the developer's registry (not just the 3 tests that happened to go red on your machine). Also added `test_legacy_installer_registry_used_as_last_resort`, which overrides that stub to a channel-suffixed path and locks the resolution order (fires only after both QSettings keys come back empty, before filesystem auto-detection). Full suite: 1091 passed.

src/utils/settings.py
Stealrull: Collapsed both registry reads into `_read_legacy_installer_sc_directory()` in a73cdb6 — get_game_install_path() and get_sc_install_root() both call the shared helper now.

src/utils/settings.py
Stealrull: Fixed in a73cdb6 — the shared helper uses `with winreg.OpenKey(...) as registry_key:`, so CloseKey always runs even on the exception path.

src/utils/settings.py
Stealrull: Fixed in a73cdb6 — the shared helper's except clause is just `except OSError:` now.

src/gui/main_window.py
Stealrull: Fixed in a73cdb6 — both call sites now use `self.statusBar().showMessage(tr("status_bar.no_strings_loaded"))`.

src/gui/main_window.py
Stealrull: Renamed to the public `robust_rmtree` in a73cdb6 at both import sites (DataForge clear-cache path and this one).

src/utils/dataforge_diff.py
Stealrull: Acknowledged — leaving as-is per your own note ("no change needed here, just adding it to the pile"), noted for the future progress-strings i18n sweep.

Osiris-DevWorks (approved):
Re-reviewed the isolation commit. This is exactly the fix: the shared json_backend fixture stubs _read_legacy_installer_sc_directory to None so nothing in this file depends on the developer machine's registry anymore, and the new test_legacy_installer_registry_used_as_last_resort locks the resolution order (after both QSettings keys, before filesystem auto-detection) with the stub overridden on purpose.

Confirmed on the machine that was failing: full suite now runs green with the real legacy SC Localization Editor tree present in HKCU (1091 passed, 16 skipped). That was the last open item on this PR.

Approving. This also unblocks #233's rebase-and-swap onto the shared win_paths helper once this lands.

PR #232, opened 2026-07-13
Feature: Mission Titles redesign (per-word checkboxes, hauling standardization)

Stealrull:
## Summary
- Replaces the old all-or-nothing "shorten original titles" toggle with individually selectable phrase/word removals (Cargo, Haul, Rank merged across dash- and comma-led contexts)
- Adds a configurable Rank separator (matching the existing Title separator options)
- Adds an Underline Direct option: renders a real Qt underline in the UI and writes the actual `<EM3>DIRECT</EM3>` tag to generated output
- Adds "Standardize hauling mission names": rewrites haul titles carrying both rank and cargo-size tokens into one canonical shape, regardless of company wording (Covalex/DeadSaints/RedWind/Ling Family)
- Splits the `[BP]`/`[ACE]` mission-title tags out of the shared mission-detail-fields toggle into an independent "General Tags" section, replacing the old non-functional "Scanning Missions" demo box
- Fixes the Tag Builder's category tabs and other filter dropdowns capturing mouse-wheel scroll instead of passing it through to the page

## Release-notes-worthy default-behavior changes (confirmed intentional)
Two changes here affect every install that never touches the Tag Builder — flagging per review feedback so these land in the 2.2.0 release notes when drafted:
- **Route placement default flips from before-title to after-title.** The `route` element defaults enabled, so an untouched install's haul/delivery titles now show the pickup→dropoff route *after* the stock title instead of before. Anyone who saved a Tag Builder config keeps their stored placement — this only affects defaults.
- **The Rank separator is always-on** (not gated by any checkbox) and normalizes comma-led stock titles to a dash (e.g. "Master Rank, Haul Cargo" -> "Master Rank - Haul Cargo"). This means an untouched install no longer reproduces every stock title byte-for-byte once mission-title enhancements are applied. `test_rank_separator_normalizes_comma_titles_by_default` pins this on purpose.

Both are intentional — confirmed with the repo owner.

## Test plan


Stealrull:
Appended 3 follow-ups: General Tags checkboxes not marking Save Tag Changes dirty, Reset-to-defaults now covering the General Tags checkboxes, and simplified General Tags hint wording. Also note: #238 and #239 are stacked on this branch — merge this one first.

src/utils/tag_builder.py
```diff
@@ -503,11 +716,12 @@ def from_json(cls, blob: str) -> "TagConfig":
...
-    # placement, not by element styles. Default: route-led (prepend), ">",
-    # " - " join, full address (2.1.1, #200: short |name can fail to resolve
-    # for some mission instances). Absorbs the #166 route-in-title toggle.
+    # placement, not by element styles. Default: route AFTER the title
+    # ("append"), ">", " - " join, full address (2.1.1, #200: short |name
+    # can fail to resolve for some mission instances). Absorbs the #166
+    # route-in-title toggle.
     elements=[ElementSpec("route", True, "")],
-    placement="prepend",
+    placement="append",
```
Osiris-DevWorks, line 724: This default flip (prepend to append) moves the route to after the title for every user still on the default config, since the route element itself defaults to enabled. Probably the right call for readability, but it's a visible change for people who never touched the Tag Builder and the PR summary doesn't mention it. Confirming this is intentional, and it should get a line in the 2.2.0 release notes either way. Users who saved a config keep their stored placement, so the blast radius is defaults-only.

src/utils/tag_builder.py
```diff
@@ -382,14 +553,33 @@ class TagConfig:
...
+    # the "shorten original titles" feature, so the route plus tags don't
+    # overflow the contract list (#200 follow-up; generalized to per-word
+    # checkboxes). Empty by default — off until the user opts in.
+    abbreviated_phrases: frozenset[str] = field(default_factory=frozenset)
+    # Separator rendered after the "Rank" position — key from RANK_SEPARATORS.
+    # Independent feature: always applied when a Rank context is detected,
+    # regardless of whether the "rank" word-removal checkbox — or any other
+    # checkbox on this page — is on. Defaults to "dash" so it's on by itself
+    # out of the box, matching what stock dash-led titles already show.
+    rank_separator: str = "dash"
```
Osiris-DevWorks, line 564: Same class of change as the placement default: with rank_separator defaulting to "dash" and the separator being always-on, a comma-led stock title ("Master Rank, ...") becomes "Master Rank - ..." with every checkbox off. The test suite pins this on purpose (test_rank_separator_normalizes_comma_titles_by_default), so I know it's deliberate, but it does mean an untouched install no longer reproduces stock titles byte-for-byte once mission-title enhancements are on. Fine if that's the product call; it just needs to be a stated one (summary + release notes).

src/utils/tag_builder.py
```diff
@@ -168,39 +169,154 @@
...
+        f"{_HAUL_RANK_TOKEN} {word}{sep} {direct}{_HAUL_SIZE_TOKEN} "
+        f"Cargo Haul{circuit}"
+    )
+
+
+# Keys covered by the legacy `abbreviate_title: true` bool migration — scoped
+# to the original shortening feature only. UNDERLINE_OPTIONS is a distinct,
+# newer feature and deliberately excluded so upgrading users don't get a
+# visual change (EM3 emphasis) they never opted into.
+_LEGACY_MIGRATION_KEYS: frozenset[str] = (
```
Osiris-DevWorks, line 303: The migration comment says legacy abbreviate_title users keep "equivalent shortening", but this key set is a superset of the old TITLE_ABBREVIATIONS map: ling_family_prefix removes a company prefix the old map never touched, and ling_family_rank inserts "Rank -" where the old pass did nothing, which can make a Ling title longer than it was. Both are probably what a shorten-everything user wants, so I'm fine shipping it; just make the comment honest ("a superset of the old behavior, including the new Ling Family handling") so nobody debugs a Ling title change back to this line expecting equivalence.

src/utils/tag_builder.py
```diff
@@ -168,39 +169,154 @@
...
+# itself always re-renders per `TagConfig.rank_separator` regardless of
+# whether "Rank" is removed (see `abbreviate_title`).
+RANK_TRIGGERS: tuple[str, ...] = (" Rank -", " Rank,")
+
+# Separator rendered after the "Rank" position, independent of whether Rank
+# itself was removed. Identical option set to TITLE_SEPARATORS (dash listed
+# first — it's the default) rather than a bespoke list: "space" collapses to
+# the same visible result a dedicated "none" option would after whitespace
+# normalization, so there's no need for a separate "none" entry.
+RANK_SEPARATORS: tuple[tuple[str, str, str], ...] = (
```
Osiris-DevWorks: This is a verbatim copy of TITLE_SEPARATORS, and test_rank_separator_matches_title_separator_options asserts they're equal, which is the test suite telling us it's one list. RANK_SEPARATORS = TITLE_SEPARATORS (plus the comment explaining why they share options) gives the same semantic name without the duplicate literals, and the test can stay as the lock.

src/gui/enhancements_tab.py
```diff
@@ -42,6 +47,29 @@ def wheelEvent(self, event):  # noqa: N802 (Qt override)
         else:
             event.ignore()
 
+    def showPopup(self):  # noqa: N802 (Qt override)
```
Osiris-DevWorks, line 50: Forcing the popup below unconditionally trades one glitch for another: for a combo near the bottom screen edge, Qt flips the list above because below would run off-screen, and this override moves it right back off-screen. Worth a guard: only force below when mapToGlobal + popup height fits the available screen geometry, else leave Qt's placement alone.

scripts/generate_enhancements_ini.py
```diff
@@ -1067,20 +1067,21 @@ def _title_has_route_token(title: str) -> bool:
...
-def _size_abbreviation_overrides(loc) -> dict:
+def _size_abbreviation_overrides(loc, shortened_sizes: frozenset = frozenset()) -> dict:
     """Loc-key overrides that shorten cargo-grade size words (#200 follow-up).
 
-    When the Shorten-original-titles toggle is on, the grade words a haul
-    title's ``~mission(CargoGradeToken)`` resolves through are abbreviated at
-    the source ("Extra Small" -> "XS") by overriding the
-    ``HaulCargo_CargoGrade_*`` / ``CargoScale_*`` loc keys. Exact value match
-    only, so an unmapped grade passes through untouched.
+    Each size is independently opted in via its own None/abbreviation
```
Osiris-DevWorks: This docstring (and the matching comments in tag_builder and the second-pass block) describe each size having "its own None/abbreviation dropdown", but the shipped UI is a single "Shorten cargo sizes" master checkbox that sets all-or-nothing. The data model supports per-size, which is good future-proofing, but the comments describe controls that don't exist. Either align the wording with the checkbox or ship the dropdowns; as-is the next reader will hunt the Enhancements tab for dropdowns they can't find.

Osiris-DevWorks (changes requested):
Full pass done. The engineering here is strong: the legacy bool migration with stale-key filtering is careful work, the one-shot General Tags migration respects prior choices, and the test coverage is the best of this batch (real stock title shapes per company, composition tests, the empty-xp crash regression). Requesting changes on two things:

1. Docs drift: HELP.md line 132 still describes the old "Shorten original titles" toggle ("trims stock-title boilerplate and abbreviates cargo-grade sizes"). This PR replaces that UI with per-word checkboxes, a Rank separator, Standardize, Underline Direct, and a General Tags box. The docs need to move with it, same PR.
2. Two default-behavior changes that aren't in the PR summary need a deliberate sign-off: route placement flips from before-title to after-title for every default-config user, and the always-on Rank separator rewrites comma-led stock titles to dashes with zero boxes ticked. Both look intentional in the tests, but they change what untouched installs see, so they belong in the summary and the 2.2.0 release notes. Details inline.

The rest inline is polish. Also noting for the record: #238 and #239 stack on this branch, so landing this first is the right order.

src/utils/tag_builder.py
Stealrull: This was an intentional change as you suspected for readability will add it to release notes.

src/utils/tag_builder.py
Stealrull: Again was intentional, I changed this thinking the previous was actually the change as default in game (no global.ini at all) uses dashes for the bulk of missions so thought it made it more new user friendly, will add to summary + release notes.

scripts/generate_enhancements_ini.py
Stealrull: This is my mistake, I originally did make the dropdown with options for each size like with components etc but decided it was over engineered (for just cargo sizes) and changed it to the current "Shorten cargo sizes", will align the wording for now with possibility of readding the individual options in the future if users request it.

src/utils/tag_builder.py
Stealrull: Fixed the comment wording in ab28d5e: "gets a superset of the old behavior — every current option key turns on, including the new Ling Family handling (`ling_family_rank`/`ling_family_prefix`) that didn't exist to opt into before."

src/utils/tag_builder.py
Stealrull: Done in ab28d5e: `RANK_SEPARATORS = TITLE_SEPARATORS` (alias) with a comment pointing at `test_rank_separator_matches_title_separator_options` as the lock.

src/gui/enhancements_tab.py
Stealrull: Added the fits-on-screen guard in ab28d5e — only forces the popup below when `mapToGlobal + popup.height()` still fits within `screen.availableGeometry()`, otherwise leaves Qt's own placement alone.

Osiris-DevWorks (approved):
Re-reviewed after ab28d5e9. Every item from the first pass is handled:

- HELP.md now describes the shipped two-checkbox UI instead of the old single toggle.
- RANK_SEPARATORS is an alias of TITLE_SEPARATORS, so the two option sets can't drift. The existing equality test stays as the lock, which is exactly the shape I hoped for.
- The migration comment now says "superset", which is honest about the Ling Family keys a legacy user gets opted into.
- The shortened_sizes comments describe the real master checkbox, not the per-size dropdowns that were never built.
- The popup override only forces "below" when it actually fits on-screen, so a combo near the bottom edge keeps Qt's flip behavior.

Full suite runs green on this branch on my machine (1098 passed, 16 skipped).

Two things to carry forward, neither blocking this PR: the placement default flip (prepend to append) and the always-on rank separator both change what an untouched install renders, so they each need a line in the 2.2.0 release notes when we draft them.

Approving. This unblocks #238 and #239, which are stacked on this branch.

Issue #222, opened 2026-07-08
Populate Blueprint Ownership from Star Citizen Log Files


Osiris-DevWorks:
Reports indicate that blueprint attribution may be signaled in SC log files, and that old log files are stored in C:\Program Files\Roberts Space Industries\StarCitizen\LIVE\LogBackups

We should see if we can parse through these files for blueprint info.
- we can ignore any log files prior to march 2026 timestamp (blueprints didn't exist before then)
- if parsing takes a long time, we should get progress bar like we do when generating enhancements
- it should be triggered by the user to execute this via a button in the blueprint section of enhancements
- we should maintain a watermark after doing this, so that we don't rescan/re-import again from older files since the last scan

Osiris-DevWorks:
this has been implemented in 2.2.0 and tested as working.

Issue #223, opened 2026-07-08
Tooltips Need Translation


Osiris-DevWorks:
All tooltips are not translated. They should be added to the ui.json language files with the human-first/ai-second translation methodology.

Stealrull:
Added in 2.2.0

PR #233, opened 2026-07-13
Feature: Blueprint Tracker tab + dirty-state/color indicators

Stealrull:
## Summary
- Splits the Blueprints shuttle out of the Enhancements tab into its own "Blueprint Tracker" tab, with a "Show tags on names" toggle and several matching/classification fixes that came with the split (ship-weapon blueprints landing in "Other", wrong blueprint-list tags, Class/Size facet display)
- Adds its own "Scan Logs for Owned Blueprints" (parses Game.log/LogBackups for blueprint reward notifications) and an "Apply Owned Tags" button, both with dirty-state tracking
- Adds dirty-state tracking + color indicators app-wide: Generate Enhancements, Save Tag Changes, Apply Enhancements, Apply Owned Tags, and the Favorites Apply Prefix button all grey out until something relevant changes, then turn red (needs applying), then green again once applied

## ⚠️ Known overlap with #222
This branch's log-scanning code (`src/utils/blueprint_log_scanner.py`, the "Scan Logs for Owned Blueprints" button) was built independently and **duplicates the already-merged "BP Scan" feature** (#222, commit f390d57) — same purpose (read Game.log for blueprint rewards), different implementation, living on a different tab. This PR keeps my version as-is rather than reconciling the two, so you can compare both and decide which to keep. I removed only the now-orphaned upstream test file (`test_blueprint_log_scan.py`, which tested code this PR's tab-split doesn't use) and one whitespace-normalization test assertion my version doesn't implement — I didn't touch any of the actual #222 feature logic.

## Test plan
- [ ] Owner review: decide how to reconcile with #222's BP Scan feature


Stealrull:
Appended 3 follow-ups: Extract from Data.p4k button text turns red when a newer Data.p4k is available, a close-time warning when Apply Enhancements has unapplied changes, and a fix for that warning's Apply Now button closing the app instead of staying open. Also note: #237 (tooltip translations) is stacked on this branch.

src/gui/workers.py
```diff
@@ -443,6 +443,36 @@ def run(self):
             self.finished.emit(False)
 
 
+class BlueprintLogScanWorker(QThread):
```
Osiris-DevWorks: This module now defines BlueprintLogScanWorker twice: the #222 original at line 233 and this one. Python takes whichever is defined last, so the old one is dead code, and it still does `from src.utils.blueprint_log_scanner import scan_channel`, which this PR deleted. If anything ever resolves the first definition (a reorder, a direct import in a test), that's an ImportError. Whichever engine wins the reconciliation, there must be exactly one class with this name when this lands.

src/utils/blueprint_log_scanner.py
```diff
@@ -1,215 +1,93 @@
...
-                            continue
-                        if latest is None or ev.timestamp > latest:
-                            latest = ev.timestamp
-                        if since is not None and ev.timestamp <= since:
-                            continue
-                        names.add(ev.name)
-                        matched += 1
-        except OSError as exc:
-            logger.warning("Blueprint log scan skipped %s: %s", path, exc)
+            text = path.read_text(encoding="utf-8", errors="replace")
```
Osiris-DevWorks: read_text loads the whole file; the #222 scanner read line-by-line with a cheap substring reject before the regex, precisely because live Game.logs run to tens of MB and logbackups can hold hundreds of files. Combined with losing the watermark and the mtime pre-filter, every scan now reads a player's entire log history from byte zero. On the deep OneDrive/Defender-intercepted setups this app already fights (see dataforge_diff's comments), that's a real regression, not a theoretical one.

src/utils/owned_items.py
```diff
@@ -56,24 +54,19 @@
...
-    plain space), removal of any ``[Owned]`` tag, removal of a leading *and* a
-    trailing bracketed component tag (``[Mil-S1-A] Norfield`` and
-    ``Barbican [IND-S3-B]`` both reduce to the bare name), and whitespace
-    collapse. Used for both the owned set and bullet matching, so a tagged
-    bullet, a log-imported name, and a bare item row all resolve to one key.
-
-    Both sides of every comparison pass through here (the owned-set entries and
-    the mission bullets in ``apply_owned_to_value``), so the folding is
-    symmetric and can never introduce a one-sided mismatch.
+    Strips a leading OR trailing component tag (``[Mil-S1-A] Norfield`` /
```
Osiris-DevWorks: This drops the NFKC unicode fold and whitespace collapse on purpose (the PR description mentions removing the assertion for it). But that normalization wasn't decorative: SC's logs emit non-breaking spaces in item names ("Lynx Legs" was the documented case), and without the fold a log-imported name silently fails to match its loc-data counterpart. Deleting the behavior and its test together is how regressions get reintroduced. Please restore both.

src/utils/pak_extractor.py
```diff
@@ -29,6 +29,30 @@
 _RMTREE_CB_KWARG = "onexc" if sys.version_info >= (3, 12) else "onerror"
 
 
+def _win_long_path(path) -> str:
```
Osiris-DevWorks: Heads up: #231 introduces this exact helper as the shared src/utils/win_paths.py, with its own test file, on this same release branch. Two sibling PRs each carrying a private copy of the same fix guarantees either a merge conflict or silent duplication depending on merge order. Since #231 lands first in the stack, rebase this on top of it and import win_long_path from win_paths instead.

src/gui/main_window.py
```diff
@@ -4722,109 +4845,93 @@ def _recompute_owned(self):
...
+        """
+        from src.utils.blueprint_log_scanner import find_log_files
+
+        channel_path = AppSettings.get_channel_install_path()
+        if not channel_path:
             QMessageBox.warning(
-                self,
-                tr("enhancements.bp_scan_title"),
-                tr("enhancements.bp_scan_no_path"),
+                self, "Scan Logs",
```
Osiris-DevWorks: The #222 version of this flow was fully translated (enhancements.bp_scan_* keys in ui.json); the replacement dialogs here are hardcoded English ("Scan Logs", the no-path warning, the results messages, the Unapplied Changes dialog, the apply tooltips). That's an i18n regression against the 2.0 language system. These need tr() keys, and the orphaned bp_scan_* keys should either be reused or removed. I know #237 is stacked on this branch for tooltips; these are dialogs, so they need handling here.

src/gui/main_window.py
```diff
@@ -1477,6 +1500,48 @@ def load_default_values(self):
         else:
             logger.debug(f"Cache file not found: {cache_file}. Default values will be empty until sources are downloaded.")
 
+    # Same enabled/disabled tooltip pattern as the Enhancements tab's
+    # Generate Enhancements / Save Tag Changes buttons.
+    _APPLY_ENABLED_TOOLTIP = (
+        "Write the merged table contents to the game's global.ini. "
+        "A timestamped backup of the current global.ini is created first."
+    )
+    _APPLY_DISABLED_TOOLTIP = (
```
Osiris-DevWorks: Wording nit for when this gets its tr() key: house style keeps em dashes out of product copy, so "Already applied — nothing has changed" should be a colon or a period. Same for the other new user-facing strings in this file.

Osiris-DevWorks (changes requested):
Full pass done. Credit where due first: the dirty-state system is genuinely well designed (single chokepoints for button state, the conservative boot default with its reasoning written down, the close-time warning that stays open after Apply Now), the new tab is cleanly signal-based, and the docs, tutorial anchors, and CLAUDE.md guides all moved with the code. That part is ready.

What blocks this is the #222 collision, and I appreciate that you flagged it honestly in the description. The answer is: reconcile, don't replace. The rewrite drops real capabilities the merged scanner had, and it leaves broken remnants behind:

1. Lost: the incremental watermark (every scan now re-reads a player's entire log history every time), the epoch floor, the mtime pre-filter, and line-by-line streaming (read_text loads tens-of-MB live logs whole). The watermark settings helpers and the test_plan.py checklist step that describes watermark behavior are now orphaned.
2. workers.py now has two classes named BlueprintLogScanWorker (lines 233 and 446). Python silently uses the second; the first still imports scan_channel, which no longer exists. That's the "didn't touch #222's logic" framing leaving a landmine.
3. normalize_item_name lost the NFKC fold + whitespace collapse that #222 added for real log names ("Lynx Legs"), and the PR deletes the test that caught it. That's a reintroduced bug, not a style choice.
4. The #222 flow was fully tr()-translated; its replacement dialogs are hardcoded English.
5. The long-path helper here duplicates the shared win_paths.py that #231 introduces on this same release branch. One of the two PRs has to own it.

My ask: keep this PR's tab, dirty-state work, and facet/matching fixes, and wire the Scan Logs button to the existing #222 scanner engine (its worker already returns raw names for the main thread to normalize and merge). If any of the old engine's behaviors genuinely should go, drop them deliberately with the test updated to match, not as a side effect of a parallel build. Details inline.

src/gui/workers.py
Stealrull: Removed the dead duplicate — kept the original (watermark-aware, calling scan_channel) and deleted the #223 regression's second definition.

src/utils/blueprint_log_scanner.py
Stealrull: Restored the #222 streaming read — scan_files now reads line-by-line with a cheap "Received Blueprint:" substring reject before the regex, no whole-file read_text.

src/utils/owned_items.py
Stealrull: Restored — normalize_item_name is back to NFKC-fold + owned-tag strip + whitespace-collapse, with test_collapses_internal_whitespace locking it in tests/test_owned_items.py.

src/utils/pak_extractor.py
Stealrull: Agreed, leaving pak_extractor.py's private _win_long_path as-is for now — this needs #231 (which introduces the shared win_paths.py) to merge first, then this branch rebases onto it. Tracking it as a blocked follow-up rather than duplicating the helper here.

src/gui/main_window.py
Stealrull: Fully i18n'd the flow — reuses the pre-existing (previously unused) enhancements.bp_scan_* keys for the scan dialog, plus new dialogs.unapplied_changes_* and toolbar.apply_btn_* keys for the rest.

src/gui/main_window.py
Stealrull: Fixed the em dash to a colon per house style: "Already applied: nothing has changed since the last Apply to Game."

Osiris-DevWorks (dismissed):
Re-reviewed after the reconciliation commit. This resolves the big one from the first pass: the scanner is back on #222's engine, and honestly it came back stronger than it left.

- BLUEPRINT_EPOCH, the watermark parameter, and the mtime pre-filter in find_log_files are all restored.
- Files stream line by line again, with a cheap "Received Blueprint:" substring reject before the regex touches anything. Whole-file read_text is gone.
- The event regex is re-anchored on the authoritative SHUDEvent_OnNotification line with a timestamp capture, so echo and continuation lines can't double-count.
- The dead duplicate BlueprintLogScanWorker is gone; the survivor uses the watermark-aware ScanResult API, and MainWindow reads and advances the watermark (monotonically, even on empty scans).
- normalize_item_name has NFKC plus whitespace-collapse back, with the test restored.
- The deleted watermark/epoch/mtime coverage from test_blueprint_log_scan.py is merged into test_blueprint_log_scanner.py.

Full suite runs green on this branch on my machine (1069 passed, 16 skipped).

The one open item is the private _win_long_path copy in pak_extractor.py, and I agree with the deferral: the shared win_paths.py lives on #231 and doesn't exist on this branch's base yet. Holding you to the stated plan, though. Once #231 merges, rebase this branch onto release/2.2.0 and swap the private copy for the shared helper before this merges. Same story for #237, which stacks on this branch and needs the rebase ripple.

Approving on that basis.

Stealrull:
Rebased onto the current release/2.2.0 tip (past #232) — one trivial import-line conflict in enhancements_tab.py, resolved by dropping the now-unused QAbstractItemView/QListWidget/QListWidgetItem imports since the blueprint shuttle code they backed already moved out to blueprint_tracker_tab.py. Full suite green (1132 passed, 16 skipped). Leaving the win_paths.py swap for pak_extractor.py's private _win_long_path until #231 lands, per your note — will do that as a follow-up commit right after.

Osiris-DevWorks (dismissed):
Re-approving after the rebase onto release/2.2.0 (post-#243). Diff is clean, full suite green on the rebased branch on my machine (1132 passed, 16 skipped).

The one standing condition carries over: pak_extractor still uses its private _win_long_path, which is correct today because the shared win_paths.py only exists on #231. #231 is now approved, so the order is: #231 merges, this branch rebases once more and swaps the private helper for src/utils/win_paths.py, then this merges, then #237 last.

PR #235, opened 2026-07-14
Feature: Medical Consumables enhancement category (CureLife pens)

Stealrull:
## Summary
- New **Medical Consumables** enhancement category: appends a curated "--- EFFECT ---" block to the 7 base CureLife pens (AdrenaPen, CorticoPen, DeconPen, DetoxPen, MedPen, OpioPen, OxyPen), whose stock descriptions are lore-only and never say what the item actually does.
- Unlike every other category this one has **no DataForge XML dependency** — `MEDICAL_CONSUMABLE_EFFECTS` is a fixed dict keyed off loc keys already in base.ini, so it's exempt from `CATEGORY_SUBTREES`/`DATAFORGE_KEEP_SUBPATHS`.
- Own checkbox on the Enhancements tab, output file `medical_consumables_enhancements.ini`, follow-up commit gives the pens their own "--- EFFECT ---" header (instead of sharing "--- STATS ---") with no redundant "Effect:" prefix.

## Testing
- `tests/test_medical_consumables.py` locks the pen set, append/prepend behavior, missing-key skip, and no-EM4 guarantee.
- Full suite green on this branch (1065 passed, 16 skipped).


scripts/generate_enhancements_ini.py
```diff
@@ -5291,6 +5298,47 @@ def _mirror_scitem_siblings(out: dict[str, str], loc: dict[str, str]) -> tuple[i
...
+    "item_Desccrlf_consumable_steroids_01":       "Reduces vision and hearing symptoms, normalizes stamina.",
+    "item_Desccrlf_consumable_radiation_01":      "Reduces injuries from radiation.",
+    "item_Desccrlf_consumable_overdoseRevival_01": "Revives an overdosed person (if not incapacitated), doubles decay rate of Blood Drug Level.",
+    "item_Desccrlf_consumable_healing_01":        "Restores health and stops bleeding. When used on another person recovers from incapacitated state.",
+    "item_Desccrlf_consumable_painkiller_01":     "Reduces pain symptoms, normalizes movement ability.",
+    "item_Desccrlf_consumable_oxygen_01":         "Recharges Oxygen reserves of a suit.",
+}
+
+
+def enhancements_medical_consumables(ctx: dict) -> dict[str, str]:
```
Osiris-DevWorks, line 5318: One subtle interaction worth a sentence in the docstring (or a follow-up): because this category has no CATEGORY_SUBTREES entry, the post-extraction dirty-category pass will never flag it, so it never regenerates automatically. Its only input is base.ini, which changes at exactly that moment, so if CIG rewords a pen's lore, the stale lore stays baked into medical_consumables_enhancements.ini until the user runs a manual full Generate. Low stakes (pen lore rarely changes), but since the whole category is an exception to the freshness system, the exception should be written down where the next person will look. Always including static categories in the post-extract regen would also fix it outright; they're nearly free to rebuild.

tests/CLAUDE.md
```diff
@@ -39,6 +39,7 @@ Split by domain:
 - `test_tag_builder.py` — TagConfig serialization + `render_tag` output shape.
 - `test_tag_config_settings.py` — TagConfig persistence via `AppSettings`. Covers the 1.4.0 rename `Phys`/`Distort`/`Bio` → `Physical`/`Distortion`/`Biochemical`.
 - `test_mining_salvage_stats.py` — 1.4.0 `enhancements_mining_laser` / `enhancements_salvage_tool` extractors. Per-mode beam stats for mining heads and handheld salvage tools, fabricated XML.
+- `test_medical_consumables.py` — 2.2.0 `enhancements_medical_consumables`, the one enhancement category with no DataForge XML dependency. Static `MEDICAL_CONSUMABLE_EFFECTS` dict driven straight off a plain `loc` dict (no XML, no tmp_path cache): every known CureLife pen key gets an appended `Effect:` line while the original lore is preserved, unknown/missing keys are skipped silently, unrelated loc keys are untouched, `stats_prepend` puts the effect line first, no `<EM4>` tags, and the locked set of exactly 7 known pen keys (base AdrenaPen/CorticoPen/DeconPen/DetoxPen/MedPen/OpioPen/OxyPen — not the Xtra variants, BoostPen, or VitalityPen).
```
Osiris-DevWorks, line 42: Tiny wording drift: this says every pen "gets an appended Effect: line", but the follow-up commit removed the Effect: prefix (the tests now assert it's absent). Say "effect line" without the colon so the blurb matches the behavior it describes.

Osiris-DevWorks (approved):
Approving. This is how a new category should land: the generator function follows the established pattern, it's registered everywhere it needs to be (jobs, phase count, workarounds loop, write gate, totals), the static-category exemption from CATEGORY_SUBTREES/DATAFORGE_KEEP_SUBPATHS is documented in all three CLAUDE.md guides plus a new reference-pattern row in the modification table, and the test file locks the pen set, the append/prepend behavior, and the no-EM4 rule. The EFFECT header instead of STATS, with no redundant prefix, reads right.

Two small notes inline, neither blocking.

tests/CLAUDE.md
Stealrull: Nice catch, my original did have Effect: but I moved Effect to a header and removed the Effect: but missed this.

PR #236, opened 2026-07-14
Feature: language selection page as the installer's first screen

Stealrull:
## Summary
- Adds a language selection page (English / Français / Português-BR / Español) as the first screen of the Inno Setup installer, writing the chosen language to the app's settings so first launch comes up already localized.
- Installer-only change (`installer.iss`); no app code touched.

## Testing
- Installer script change — verify by compiling `installer.iss` with Inno Setup and stepping through the wizard (page order, each language choice persisting into the app's `SELECTED_LANGUAGE` setting).

### DIFF
--- a/installer.iss
+++ b/installer.iss
@@ -75,6 +75,14 @@ var
     HKCU\...\ui_mode in WriteInstallerChoicesToRegistry and the app reads it on
     first launch. Defaults to Simple, pre-selected from a prior install. }
   ModeChoicePage: TInputOptionWizardPage;
+  { App UI/string language. Radio page anchored to wpWelcome so it's the very
+    first page the user sees, before any of the built-in wizard pages. The
+    choice is written to HKCU\...\selected_language (matching AppSettings.
+    SELECTED_LANGUAGE) so the app opens in that language on first launch
+    instead of defaulting to English until the user finds the Config tab's
+    language selector. Options must stay in sync with the non-stub folders
+    under languages/ (AppSettings.get_available_languages()). }
+  LanguageChoicePage: TInputOptionWizardPage;
 
 function IsAutoUpdate(): Boolean;
 begin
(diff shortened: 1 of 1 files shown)


installer.iss
```diff
@@ -703,6 +711,23 @@ begin
...
+  case LanguageChoicePage.SelectedValueIndex of
+    1: RegWriteStringValue(HKCU,
+         'Software\Osiris DevWorks\Smart Citizen', 'selected_language', 'french');
+    2: RegWriteStringValue(HKCU,
+         'Software\Osiris DevWorks\Smart Citizen', 'selected_language', 'portuguese_br');
+    3: RegWriteStringValue(HKCU,
+         'Software\Osiris DevWorks\Smart Citizen', 'selected_language', 'spanish');
+  else
+    RegWriteStringValue(HKCU,
+      'Software\Osiris DevWorks\Smart Citizen', 'selected_language', 'english');
```
Osiris-DevWorks, line 728: There's a future trap in the always-write plus the unknown-value fallback. The day a fifth language ships in the app but this page's list lags behind (three manually-synced lists now: the page options, this case map, and languages/), an upgrade install will read the user's saved value, fail to map it, pre-select English, and this block will overwrite their choice with 'english'. Silent installs through the auto-update chain hit the same path. Three-line guard: if RegQueryStringValue found a saved value that didn't map to any option index, skip the write entirely and leave the unknown value alone. Then the worst case for a lagging installer is "page shows English pre-selected" instead of "user's language got reset".

Osiris-DevWorks (approved):
Approving. I verified the load-bearing details: the registry value name matches AppSettings.SELECTED_LANGUAGE exactly, the four written values match the languages/ folder names exactly, the reinstall pre-select keeps an upgrade from resetting a non-English user, and the ASCII-labels rationale (no BOM in installer.iss, ANSI codepage risk) is the right call and written down. One hardening ask inline; it can't bite today, so it doesn't block.

installer.iss
Stealrull: I wasn't sure how this would handle a new language being added one day, glad to have you, adding the guard.

PR #237, opened 2026-07-14
Feature: translate every hardcoded UI tooltip (EN/FR/PT-BR/ES)

Stealrull:
## Summary
- Moves every hardcoded English tooltip across the app (Config, Enhancements, Blueprint Tracker, main window, filter header, import dialog, log tab, tag mapping dialog) into the `tr()` i18n system, with ~292 new keys per language translated into French, Portuguese-BR, and Spanish.

## Stacking note — merge this one LAST
The tooltip pass covers widgets introduced by the other open PRs, so this branch is **stacked on #232 (Mission Titles redesign), #233 (Blueprint Tracker), and #235 (Medical Consumables)** — their commits appear in this PR's diff until they merge. Merge those three first and this PR's diff collapses to just the tooltip/i18n work.

## Testing
- Full suite green on this branch (1102 passed, 16 skipped); all four `ui.json` files parse; app launch smoke test clean.

### DIFF
--- a/src/gui/filter_header.py
+++ b/src/gui/filter_header.py
@@ -86,7 +86,7 @@ class FilterHeaderView(QHeaderView):
                 continue
             editor = QLineEdit(self)
             editor.setObjectName("columnFilter")
-            editor.setPlaceholderText(f"Filter {name.lower()}…")
+            editor.setPlaceholderText(tr("strings_tab.column_filter_placeholder", column=name.lower()))
             editor.setToolTip(tr("strings_tab.column_filter_tooltip", column=name))
             editor.setClearButtonEnabled(True)
             editor.addAction(self._search_icon, QLineEdit.ActionPosition.LeadingPosition)
@@ -110,7 +110,7 @@ class FilterHeaderView(QHeaderView):
             if editor is None:
                 continue
             name = names[i] if i < len(names) else ""
-            editor.setPlaceholderText(f"Filter {name.lower()}…")
+            editor.setPlaceholderText(tr("strings_tab.column_filter_placeholder", column=name.lower()))
             editor.setToolTip(tr("strings_tab.column_filter_tooltip", column=name))
 
     def clear_all(self):
--- a/src/gui/filter_header.py
+++ b/src/gui/filter_header.py
@@ -4,6 +4,8 @@ from PyQt6.QtCore import Qt, QTimer, QEvent, pyqtSignal, QSize, QRect, QPoint
 from PyQt6.QtGui import QColor, QIcon, QPainter, QPalette, QPen, QPixmap
 from PyQt6.QtWidgets import QHeaderView, QLineEdit, QStyleOptionHeader, QStyle
 
+from src.utils.i18n import tr
+
 
 # Resting-border alpha used to approximate a 1.5px-feeling stroke without
 # leaving the 2px integer grid that QSS borders require. ~60% opacity reads
@@ -85,9 +87,7 @@ class FilterHeaderView(QHeaderView):
             editor = QLineEdit(self)
             editor.setObjectName("columnFilter")
             editor.setPlaceholderText(f"Filter {name.lower()}…")
-            editor.setToolTip(
-                f"Filter the {name} column. Filters across columns combine with AND."
-            )
+            editor.setToolTip(tr("strings_tab.column_filter_tooltip", column=name))
             editor.setClearButtonEnabled(True)
             editor.addAction(self._search_icon, QLineEdit.ActionPosition.LeadingPosition)
             editor.textChanged.connect(self._on_text_changed)
@@ -111,6 +111,7 @@ class FilterHeaderView(QHeaderView):
                 continue
             name = names[i] if i < len(names) else ""
             editor.setPlaceholderText(f"Filter {name.lower()}…")
+            editor.setToolTip(tr("strings_tab.column_filter_tooltip", column=name))
 
     def clear_all(self):
         """Clear every filter input without triggering per-keystroke signals."""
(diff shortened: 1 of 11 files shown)


src/gui/filter_header.py
```diff
@@ -85,9 +87,7 @@ def __init__(self, column_names: list[str], parent=None, skip_columns: set[int]
             editor = QLineEdit(self)
             editor.setObjectName("columnFilter")
             editor.setPlaceholderText(f"Filter {name.lower()}…")
```
Osiris-DevWorks: The tooltip on these editors is translated now, but the placeholder two lines up (f"Filter {name.lower()}…") is still hardcoded English, in both this constructor and update_column_names. Same widget, half-translated. Worth a strings_tab.column_filter_placeholder key in the same pass, since a French user sees the placeholder far more often than the tooltip.

Osiris-DevWorks (commented):
Reviewed the tooltip/i18n work in isolation from the stacked #232/#233/#235 commits (those got their own reviews). The core of this PR is in good shape: every swap I checked routes through tr() correctly, retranslate_ui coverage was extended to the new tooltip keys (log_tab, filter_header's rename path), and all 292 new FR/PT-BR/ES entries are correctly filed as AI fallbacks (ht empty, at filled), which is exactly what the inline provenance system in TRANSLATIONS.md asks for. Nobody's human strings were touched. Nice discipline.

Holding the approve until the stack beneath it settles, for three reasons:

1. This branch includes #233's commits, and #233 has changes requested (including a rework of the log-scanner reconciliation). When that lands, this branch rebases and some keys here (blueprint_tracker.*, the apply tooltips) may shift with it.
2. Process check, your call: the translation policy has AI backfill happening at the pre-release docs-sync, not mid-integration. The provenance structure makes this safe either way (the at fields are exactly what a later human pass reviews), so I'm fine with it, but flagging since it front-loads work the policy schedules for later.
3. The English source for apply_disabled_tooltip carries the em dash I flagged on #233; fixing it there means fixing the key and its three translations here too.

One small consistency note inline. Ping me after the rebase and this should be a quick approve.

src/gui/filter_header.py
Stealrull: Just replying to the three things here.
1. Should be good with #233 now, did the rework to your log scanner, nice work on it I didn't think of the watermark etc so it doesn't have to scan all logs everytime.
2. I feel like we should just use AI as we go, in general going forward I don't think there will be extensive changes (I know I did alot this time) but feels fine to just do it as we go even if occasionally it may need to be done twice I don't mind doing it.
3. Hopefully done right.

Osiris-DevWorks (dismissed):
Re-reviewed the two follow-up commits.

- The column-filter placeholder is a real catch: the tooltip was translated but the placeholder two lines up was not, and the placeholder is what users actually stare at. Now both route through tr(), at both call sites.
- The bp_scan_* restore checks out. I verified those 7 keys only ever existed in English (the other languages never had them), so nothing human-translated was lost, and the two superseded button keys staying gone is correct since blueprint_tracker.scan_logs_btn replaced them in the tab split.

Full suite runs green on this branch on my machine (1102 passed, 16 skipped).

The verdict from the first pass stands, just upgraded: the content is ready, and the only thing between this and merge is sequencing. This branch carries #232, #233, and #235's commits, so it merges last. Once #232 and #233 land on release/2.2.0 (and #233 picks up the win_paths swap on its own rebase), rebase this on the release branch so the diff collapses to just the tooltip work.

Approving on that basis.

Stealrull:
Rebased onto the current release/2.2.0 tip (now carrying the rebased #233). Same trivial enhancements_tab.py import conflict as #233, same resolution. Full suite green (1120 passed, 16 skipped).

Osiris-DevWorks (changes requested):
Pulling my approval on this one: the restack dropped part of its base and quietly resurrected the scanner regression #233 just fixed.

What I found on the current head (d5dfa90b):

- The branch carries #233's commits rebased, but only up through "Fix Apply Now closing the app". Two of #233's commits are missing: the big reconcile commit (the one that restored BLUEPRINT_EPOCH, the watermark, mtime pre-filtering, streaming reads, and removed the duplicate BlueprintLogScanWorker) and the em dash tooltip fix.
- The code confirms it: blueprint_log_scanner.py on this branch has zero mentions of BLUEPRINT_EPOCH or watermark, the old bare "Received Blueprint:" regex, whole-file read_text, and workers.py has two BlueprintLogScanWorker classes again.
- The full suite runs green on this branch, which is the trap: the old tests came back along with the old code, so nothing red flags it. If this merges after #233, the two branches no longer share history (both rebased independently), so the merge either conflicts across all those files or, where only this branch appears to have touched them, quietly ships the regressed versions.

The fix is to restore the stacking instead of the flat rebase. This PR's own work is exactly three commits: 2dc523d5 (tooltip sweep), 5074c479 (column-filter placeholder), d5dfa90b (bp_scan_* key restore). Something like:

    git rebase --onto origin/feature/blueprint-tracker-and-apply-state cf1a6e6d feature/ui-tooltip-translations

(i.e. replay just those three onto #233's current head), then force-push. After that the PR diff vs #233 collapses back to tooltip work only, and once #233 merges you rebase onto release/2.2.0 one final time as planned.

One thing to re-check after the restack: d5dfa90b restores the 7 bp_scan_* keys, and the reconciled #233 also touches the enhancements.bp_scan_* usage, so make sure the key set still matches the tr() calls on the reconciled code (the earlier version of this branch had that right; it just needs re-verifying against the new base).

The tooltip work itself is still approved in spirit. This is purely about the base it's sitting on.

PR #238, opened 2026-07-14
Feature: Resource Signature (RS) tags for Battaglia Scan/Mining mission titles (4.9)

Stealrull:
## Summary
- New **Scanning/Mining Missions** box in the Mission Titles tag builder with a "Show Resource Signature (RS) Tags" checkbox (on by default).
- Recco Battaglia's 4.9 Scan/ScanMine contracts each target 1–3 specific mineable ores (`ResourceType`/`ResourceType2`/`ResourceType3` MissionProperty overrides). When enabled, titles gain e.g. `Full Area Scan for Ore [BP] [300 Rep] [RS 3900/3600/4300]` so miners know the scanner strength needed before accepting.
- CIG doesn't expose the RS number anywhere in DataForge (verified against the full PTU 4.9 cache), so `MINEABLE_RS_VALUES` is a curated 8-ore table modeled on the community loc reference (MrKraken/StarStrings `ptu/4.9` `mining.ini`).
- Verified end-to-end against a real PTU 4.9 DataForge extraction: all 15 Scan/ScanMine titles tag correctly, zero leakage onto Battaglia's other mission types or other contractors.

## Stacking note
Stacked on **#232 (Mission Titles redesign)** — the General Tags system this builds on. Its commits appear in this diff until #232 merges. Also touches the same `MISSION_TITLE_TAG_KEYS`/title-tag block as **#239 (reputation track tags)**; whichever merges second will need a trivial conflict resolution (the combined intended state is both keys present).

## Testing
- `tests/test_battaglia_mineable_rs.py` (15 tests): ore extraction/order/dedup, MineableType-token exclusion, tag formatting, title-prefix scoping, intro-contract coverage.
- Full suite green on this branch (1113 passed, 16 skipped).


scripts/generate_enhancements_ini.py
```diff
@@ -3103,6 +3104,100 @@ def _build_template_lookup(
...
+        text_id = opt.get("textId", "")
+        if not text_id.startswith("@mineabletype_primary_"):
+            continue
+        ore = text_id[len("@mineabletype_primary_"):]
+        if ore not in ores:
+            ores.append(ore)
+    return ores
+
+
+def _format_rs_tag(ores: list[str]) -> str:
```
Osiris-DevWorks, line 3162: Unknown ores dropping silently is the right behavior for the tag (tested, good), but it also means table drift is invisible: when CIG adds a ninth ore to this mission family, titles quietly show fewer RS values than ores and nobody finds out until a player notices. One logger.info naming the unmatched ore key here would make the next 4.x data bump self-reporting.

src/gui/enhancements_tab.py
```diff
@@ -1545,6 +1660,76 @@ def _combo(items) -> QComboBox:
...
+        tags_col.addStretch()
+        page.addWidget(tags_group)
+
+        # "Scanning/Mining Missions" (4.9+): its own box next to General Tags
+        # rather than folded into it, since it's specific to one mission
+        # family (Recco Battaglia's Scan/Mining contracts) rather than a
+        # tag shown across every mission type. Same immediate-persist
+        # pattern as the General Tags checkboxes above (shares
+        # self._title_tag_checkboxes so Reset-to-defaults picks it up too).
+        scanning_group = QGroupBox("Scanning/Mining Missions")
```
Osiris-DevWorks, line 1718: This box's three strings (group title, hint, checkbox label) are hardcoded English on a tab that's otherwise tr()-driven. Same is true of #232's General Tags labels next door, so I'd fix them together: either give both boxes ui.json keys now, or fold them into #237's translation pass so the FR/PT-BR/ES files pick them up in one sweep. As-is these are new permanently-English chrome for non-English users.

Osiris-DevWorks (dismissed):
Approving the RS-tag work (reviewed as the diff on top of #232; the stacked commits got their own review there). This is a well-scoped feature: the pre-pass follows the established _build_template_lookup shape, the curated-table justification is documented in three places with the source named (checked DataForge for the values, found none, so the community reference is the honest fallback), and the 15-test file covers the traps that matter: the MineableType/ResourceType token family split, the intro <Contract> wrapper, order preservation, dedup, and title-prefix scoping. The stacking and #239-conflict notes in the description are exactly the right transparency.

Merge order stands as you noted: #232 first, and #232 currently has changes requested, so this waits on that. Two small notes inline.

scripts/generate_enhancements_ini.py
Stealrull: Added diagnostic logging to _format_rs_tag in 02bfe88 — logs unmatched ore keys via logger.info so a future CIG data bump that adds a 9th ore is self-reporting instead of silently dropping.

src/gui/enhancements_tab.py
Stealrull: Agreed, holding off here — folding this box's strings into #237's translation sweep once #232 merges (same plan as #232's General Tags labels next door), rather than duplicating a partial i18n pass now.

Osiris-DevWorks (dismissed):
Looked at the follow-up commit. The unmatched-ore logging is exactly what I was after: the curated table now self-reports when a CIG data bump adds a mineable it doesn't know, instead of quietly rendering fewer RS values than ores. Full suite is green on this branch on my machine (1113 passed, 16 skipped).

Re-approving on the new head. Merge order note still applies: this branch carries #232's commits, so it waits for #232 to land on release/2.2.0 first, then rebases (or merges) so the diff collapses to just the RS tag work. Whichever of this and #239 goes second resolves the small MISSION_TITLE_TAG_KEYS conflict; the combined tuple should end up ("rep", "blueprint", "ace", "rs", "rep_track").

PR #239, opened 2026-07-14
Feature: show which reputation track a mission's XP feeds (#161)

Stealrull:
## Summary
Implements #161. Some factions have multiple reputation tracks (Foxwell Enforcement shows separate "Security" and "Standing" columns in-game), and which one a mission feeds isn't guessable — the ship-combat "Security Patrol" contracts feed the generic **Contractor** track while the satellite-scan "Destroy Data Skimmers" contract feeds **Security** (verified against the LIVE cache).

- New `_build_standing_tracks` lookup resolves each contract's `minStanding` rank UUID to its track name via the `RepStanding_`/`RepScope_` loc-key family (the same labels the in-game Reputation Manager uses).
- **Mission body**: the Reputation line becomes e.g. `Applicant: +500 Rep (Security)` — always on, suppressed when the rank name already equals the track (no `Contractor: +200 Rep (Contractor)` noise).
- **Mission title**: the Rep tag can become `[500 Rep (Security)]` behind a new **off-by-default** "Rep Track Tag" checkbox in General Tags, shown only when every variant of a title agrees on one track. Reset-to-defaults now honors per-field defaults.

## Stacking note
Stacked on **#232 (Mission Titles redesign)** — its commits appear in this diff until it merges. Touches the same `MISSION_TITLE_TAG_KEYS`/title-tag block as **#238 (Battaglia RS tags)**; whichever merges second needs a trivial conflict resolution (intended combined state: both keys present, `("rep", "blueprint", "ace", "rs", "rep_track")`).

## Release-notes-worthy visibility change (confirmed intentional)
The mission-body Reputation line's track suffix is unconditional (not gated by any checkbox) — flagging per review feedback so this lands in the 2.2.0 release notes when drafted: every user's MISSION DETAILS Reputation line changes the next time they generate (e.g. "Applicant: +500 Rep (Security)"), even if they never touch the new Rep Track Tag checkbox. Intentional — the body is where precision belongs; confirmed with the repo owner. The title tag is the opt-in, off-by-default part.

## Testing
- `tests/test_reputation_track.py` (track-token regex), new `TestReputationTrackSuffix` class in `test_mission_rep_label.py`, 12-field variant-tuple lock in `test_mission_variant_tuple.py`, off-by-default + gating coverage in `test_mission_title_tags.py`.
- Full suite green on this branch (1116 passed, 16 skipped). Verified end-to-end against the real LIVE DataForge cache.


scripts/generate_enhancements_ini.py
```diff
@@ -3458,12 +3478,16 @@ def _extract_system(name: str, fallback: str) -> str:
...
+                            # generic "Standing"/Contractor track — some
+                            # factions have both; see issue #161).
                             rank_name = standings_lookup.get(min_standing, "") if min_standing else ""
+                            rep_track = standing_track_lookup.get(min_standing, "") if min_standing else ""
 
                             # Add all missions (not just those with XP/blueprint data)
                             if title_key not in missions:
                                 missions[title_key] = []
-                            missions[title_key].append((system_name, success_xp, failure_xp, desc_key, contract_flags, spawns, contract_difficulty, contract_has_bp, contract_bp_chance, contract_bp_variant, rank_name))
+                            missions[title_key].append((system_name, success_xp, failure_xp, desc_key, contract_flags, spawns, contract_difficulty, contract_has_bp, contract_bp_chance, contract_bp_variant, rank_name, rep_track))
```
Osiris-DevWorks: Not this PR's fault, but this tuple just hit 12 positional fields, and downstream code reads it as v[7] and eleven-name unpacks. The next field added is the one that introduces an off-by-one nobody catches in review. Whenever someone next touches this scan, a NamedTuple (variant.success_xp, variant.rep_track) would make every one of these sites self-checking. Worth a follow-up issue so it doesn't get lost.

scripts/generate_enhancements_ini.py
```diff
@@ -6301,6 +6364,53 @@ def _builder() -> dict[str, str]:
     if _want("mission_rewards"):
         lookup_jobs["standings"] = _build_standings
 
+    def _build_standing_tracks():
```
Osiris-DevWorks: This walks and parses records/reputation/standings twice per cache build: _build_standings reads the same files to map UUID to rank name, and this reads them again for UUID to track. Both are derived from the same displayName attribute, so one pass returning both dicts (or _build_standings returning a richer value) halves the XML work and keeps the two lookups impossible to desync. The _cached_lookup wrapper makes the cost a cache-miss-only concern, so this is a tidiness ask, not a perf emergency.

scripts/generate_enhancements_ini.py
```diff
@@ -2681,7 +2682,7 @@ def _extract_difficulty(element: ET.Element) -> str:
     return ""
 
 
-def _rep_reward_line(field_name: str, amount_str: str, rep_xp_label: str) -> str:
+def _rep_reward_line(field_name: str, amount_str: str, rep_xp_label: str, track: str = "") -> str:
```
Osiris-DevWorks, line 2752: Flagging the visibility change for the record: the body-line track suffix is unconditional, so every user's MISSION DETAILS Reputation line changes the next time they generate (e.g. "Applicant: +500 Rep (Security)"). I think always-on is right here, the body is where precision belongs, but it should get a line in the 2.2.0 release notes since it changes output for people who touch nothing.

Osiris-DevWorks (dismissed):
Approving (reviewed as the diff on top of #232; the stacked commits got their own review there). This closes #161 the right way: the track comes from the same RepStanding_/RepScope_ loc-key family the in-game Reputation Manager uses, the title tag is conservative (only when every nonzero tier agrees on one track, and off by default), and two details show real care: the xp_tag_re guard was updated so the optional "(Track)" can't trick the pu-only pass into double-tagging, and Reset-to-defaults now restores per-field defaults instead of assuming everything defaults on. The #238 conflict note with the intended combined key order is exactly what the second merger needs.

Against #161's ask: mission body shows the track, title tag optional, no wrong-track guessing. Satisfied. Three non-blocking notes inline, plus one carry-over: the new checkbox label is hardcoded English like the rest of the General Tags box; fix them together per my comment on #238.

scripts/generate_enhancements_ini.py
Stealrull: Will make a follow up issue and see about fixing it as well, nice catch.

scripts/generate_enhancements_ini.py
Stealrull: Merged into one XML walk in c1902d4 — `_build_standings` now returns both the rank and track dicts from a single pass, mirroring the `_build_scitem_pair` pattern.

scripts/generate_enhancements_ini.py
Stealrull: Done in e3afe19, bundled into this PR since it's what has the 12-field shape that triggered this: added a `ContractVariant` NamedTuple and switched every `v[N]` read/destructure to attribute access (`v.desc_key`, `v.has_bp`, `v.rep_track`, etc). Still a plain tuple under the hood so `tests/test_mission_variant_tuple.py`'s positional-unpack/indexing locks needed no changes. Verified against the real DataForge cache too — output is unchanged. Closes #240.

Osiris-DevWorks (dismissed):
Reviewed both follow-up commits.

- The merged XML walk is the right shape: one _cached_lookup job now returns both the rank-name and track views of displayName instead of parsing every standings XML twice.
- The ContractVariant NamedTuple is a bigger win than the note that prompted it. Twelve positional fields read by name at every site, with the tuple contract locked by the existing test file untouched, means the next field addition can't silently shift indices. Good call splitting it out as #240 and closing it here.

Full suite is green on this branch on my machine (1116 passed, 16 skipped).

Re-approving on the new head. Same sequencing as #238: this waits for #232 to land, then rebases so the diff collapses to the rep-track work. Whichever of #238 and this merges second picks up the one-line MISSION_TITLE_TAG_KEYS conflict; combined tuple should be ("rep", "blueprint", "ace", "rs", "rep_track").

scripts/generate_enhancements_ini.py
Stealrull: Agreed, always-on is the right call here. Noting this for the 2.2.0 release notes when we do the pre-release docs pass rather than drafting them now mid-integration.

Stealrull:
Rebased again onto release/2.2.0 (past #238 and #243, which landed after the last rebase). Two conflicts, both straightforward:\n- settings.py: combined MISSION_TITLE_TAG_KEYS to ("rep", "blueprint", "ace", "rs", "rep_track") per your merge-order note, keeping both #238's rs entry and this PR's rep_track entry (+ its default-off setting).\n- generate_enhancements_ini.py: two independent additions (this PR's _REP_TRACK_PREFIX_RE block and #238's Battaglia RS tag block) landed adjacent to each other; kept both, plus combined the multi-XP title-tag line so it appends both the rep-track suffix and the RS tag. Full suite green (1140 passed, 16 skipped).

Osiris-DevWorks (dismissed):
Re-approving after the rebase onto release/2.2.0. The diff now collapses to just the rep-track work as intended, and the MISSION_TITLE_TAG_KEYS conflict against the merged #238 resolved to exactly the combined tuple we agreed on: ("rep", "blueprint", "ace", "rs", "rep_track"). Full suite green on the rebased branch on my machine (1140 passed, 16 skipped). Ready to merge.

PR #242, opened 2026-07-15
Fix: guard installer language write-back against unrecognized saved value

Stealrull:
Follow-up on [#236's review note](https://github.com/Osiris-DevWorks/smart-citizen/pull/236#discussion_r3583770395):

> There's a future trap in the always-write plus the unknown-value fallback. The day a fifth language ships in the app but this page's list lags behind... an upgrade install will read the user's saved value, fail to map it, pre-select English, and this block will overwrite their choice with 'english'. Silent installs through the auto-update chain hit the same path.

## Fix

Added a `LanguageChoiceUnknownSaved` flag: when the read side finds a saved `selected_language` value that doesn't match any of this page's known options (and isn't `'english'` either), it flags it instead of just silently defaulting to English. The write side then skips the write-back entirely when that flag is set, leaving the real saved value alone.

Worst case with the guard: the page shows English pre-selected for that one install, but the actual saved value in the registry survives untouched — matching Osiris's suggested behavior exactly.

## Test plan
- [x] `pytest tests/` — 1065 passed, 16 skipped (installer.iss has no Python test coverage; this is Inno Setup Pascal Script)
- [x] `ISCC.exe installer.iss` compiles clean end-to-end (installed Inno Setup 6.7.3 to verify) — the `[Code]` section including the new `LanguageChoiceUnknownSaved` guard parses/compiles with no errors, and a full compile against a stubbed `dist/SmartCitizen/` produces a valid Setup.exe
- [ ] Manual: install with a real build and a stubbed unrecognized `selected_language` registry value, confirm the value survives after setup runs — worth a real run before merge since I only verified the script compiles, not the runtime registry behavior end-to-end (but I'm unsure how to do this if I'm honest).

### DIFF
--- a/installer.iss
+++ b/installer.iss
@@ -726,16 +726,22 @@ begin
     the chosen language rather than relying on the app's English fallback.
     Values here must stay in sync with the folder names under languages/.
 
-    Exception: skip the write entirely when the saved value didn't match any
-    option this page knows about (LanguageChoiceUnknownSaved) — the page
-    falls back to displaying English in that case, and an always-write would
+    Exception: skip the write when the saved value didn't match any option
+    this page knows about (LanguageChoiceUnknownSaved) AND the selection is
+    still sitting at the pre-selected default (index 0) — the page falls
+    back to displaying English in that case, and an always-write would
     silently downgrade the user's real (just-unrecognized) language back to
     English, both on an interactive upgrade and on a silent auto-update
-    install where nobody sees the page to correct it. Worst case with the
-    guard: the page shows English pre-selected but the saved value is left
-    untouched. }
-  if LanguageChoiceUnknownSaved then
-    Log('Skipped rewriting selected_language: saved value matched no known option, left unchanged.')
+    install where nobody sees the page to correct it. The index-0 check
+    matters: on an INTERACTIVE upgrade the user still sees the page and can
+    actively move the selection off English to a real choice, and that
+    explicit pick must win — the flag only describes the pre-selection
+    state, not anything the user did afterward. Worst case with the guard:
+    a user who deliberately re-picks English while their saved value is
+    unknown stays on the unknown value (the safe direction — they can
+    switch in-app, and this never destroys a value it doesn't understand). }
+  if LanguageChoiceUnknownSaved and (LanguageChoicePage.SelectedValueIndex = 0) then
+    Log('Skipped rewriting selected_language: saved value matched no known option and selection was left at the default, left unchanged.')
   else
   begin
     case LanguageChoicePage.SelectedValueIndex of
--- a/installer.iss
+++ b/installer.iss
@@ -83,6 +83,15 @@ var
     language selector. Options must stay in sync with the non-stub folders
     under languages/ (AppSettings.get_available_languages()). }
   LanguageChoicePage: TInputOptionWizardPage;
+  { True when a saved selected_language value was found but didn't match any
+    of the known LanguageChoicePage options below. Guards the always-write in
+    WriteInstallerChoicesToRegistry: without it, an upgrade install whose
+    saved value has fallen out of sync with this page's option list (e.g. a
+    5th language shipped in the app but not added here yet) would silently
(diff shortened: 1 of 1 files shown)


installer.iss
```diff
@@ -715,19 +724,33 @@ begin
...
+
+    Exception: skip the write entirely when the saved value didn't match any
+    option this page knows about (LanguageChoiceUnknownSaved) — the page
+    falls back to displaying English in that case, and an always-write would
+    silently downgrade the user's real (just-unrecognized) language back to
+    English, both on an interactive upgrade and on a silent auto-update
+    install where nobody sees the page to correct it. Worst case with the
+    guard: the page shows English pre-selected but the saved value is left
+    untouched. }
+  if LanguageChoiceUnknownSaved then
```
Osiris-DevWorks: This skip fires on the flag alone, so it also swallows a deliberate pick. Walk the interactive upgrade: saved value is (say) 'german' from a newer app, the flag sets, the page shows English pre-selected, and the user actively clicks French. Write-back skips anyway, so their explicit choice on this very page gets ignored and the install exits still on 'german'.

The flag only describes the pre-selection state. Once the user moves the selection off the default, the ambiguity is gone. One-line fix:

```
if LanguageChoiceUnknownSaved and (LanguageChoicePage.SelectedValueIndex = 0) then
```

That keeps the guard for the silent install and the untouched-default case (index 0 could mean "never saw the page" or "left it alone", so skipping is right), and honors any explicit non-English pick. The one residual corner, a user who deliberately re-picks English while their saved value is unknown, stays on their unknown language, which is the safe direction: they can switch in-app, and we never destroy a value we don't understand.

For the manual check in your test plan, no real build needed:
1. `reg add "HKCU\Software\Osiris DevWorks\Smart Citizen" /v selected_language /d german /f`
2. Run your stub-compiled Setup.exe with `/SILENT`, then `reg query` the value: it should still say german.
3. Run it again interactively, pick French on the page, and the value should now say french (this is the case the current code gets wrong).
4. `reg add ... /d french /f`, run silent again: should stay french (known value, normal write path).

Osiris-DevWorks (changes requested):
Reviewed. The guard is the right instinct and the silent-install path is exactly right: unknown saved value, page never shown, write-back skipped, real value survives. The doc comments explaining the why are good, and I appreciate the honest test plan.

One gap in the interactive path before this merges, details inline. There's also a straightforward way to run the manual check you flagged, in the same comment.

installer.iss
Stealrull: Fixed in 4472645 — the guard now only skips the write when the selection is also still sitting at index 0:

```
if LanguageChoiceUnknownSaved and (LanguageChoicePage.SelectedValueIndex = 0) then
```

Ran your reg-based manual steps to verify the automatable parts:
1. `reg add ... /d german /f` then silent install → `german` survives unchanged (correct — page never shown).
2. Set to `french` (a known value), silent install → `french` still writes/persists normally through the standard path (sanity check the guard doesn't over-trigger).

I don't have a way to click through the actual interactive wizard page in my environment, so step 3 (actively picking French on the page while the saved value is `german`) isn't verified end-to-end by me — the code change is straightforward enough that I'm confident in it (once `SelectedValueIndex` moves off 0 the guard condition is false and the normal write branch fires), but flagging that this one path is reasoned-through + compiled, not click-tested.

Osiris-DevWorks (approved):
Re-reviewed the follow-up commit. The condition is exactly right: the skip now requires both the unknown-saved flag and an untouched default selection, so silent installs and left-alone pages keep protecting the unrecognized value while an explicit pick on the page wins. The rewritten comment walks the reasoning cleanly, including the deliberate re-pick-English corner staying on the safe side.

Thanks for running the reg-based checks and being straight about what they cover. The remaining gap (a real interactive click-through of the wizard page) is a two-minute test with any preview installer build: set the registry value to something fake, run the installer, pick French on the page, and confirm the registry says french afterward. Worth adding to the tester checklist for the next preview build rather than holding this PR on it, since the silent path (the dangerous one) is verified.

Approving.

PR #243, opened 2026-07-15
Feature: Base Resource Signature values in the Mining Compendium journal

Stealrull:
Stacks on #238 (feature/battaglia-rs-mission-tags) since it extends that PR's `MINEABLE_RS_VALUES` table — review as the diff on top of #238; #238's own commits already have their review there.

## What

The "Mining Fundamentals #2: Where to Mine" journal entry (`Journal_General_Mining_Compendium_Content`) already gets reformatted per-mineral into `<EM3>Name</EM3>` / `Locations:` / `Used To Craft:` blocks by `scan_crafting_blueprints`. Each ore's block now also gets a `Base Resource Signature: <EM4>{value}</EM4>` line when a confirmed value exists.

## RS value sourcing

`MINEABLE_RS_VALUES` (previously 8 ores, curated for #238's Battaglia mission tag) is expanded to **26 ores**, cross-checked against [starminersdepot.com](https://starminersdepot.com/scanner)'s Alpha 4.8 scanner table:
- 7 of the original 8 values matched that reference exactly (Aluminium, Bexalite, Iron, Lindinium, Savrillium, Titanium, Torite).
- "Ice" isn't listed on that reference at all, but is kept at its existing confirmed value.
- 18 new ore values added from the same reference (Agricium, Aslarite, Beryl, Borase, Copper, Corundum, Gold, Hephaestanite, Laranite, Ouratite, Quantainium, Quartz, Riccite, Silicon, Stileron, Taranite, Tin, Tungsten).

That covers 26 of the journal's 35 ores. The remaining 9 (Aphorite, Beradon, Caranite, Dolivine, Feynmaline, Glacosite, Hadanite, Janalite, Sadaryx) have no known RS value from any source and are left unchanged — no placeholder, no guess.

One spelling note: the journal's own prose has a CIG typo ("Savrilium", single L) that doesn't match the DataForge loc-key convention (`mineabletype_primary_savrillium`). `_MINING_COMPENDIUM_ORE_ALIASES` bridges the two so the lookup still matches.

## Test plan
- [x] `pytest tests/` — 1116 passed, 16 skipped
- [x] New `TestMiningCompendiumRSTagging` class in `tests/test_commodity_journal.py`: known ore gets the RS line, unknown ore doesn't, "Savrilium" typo-spelling still resolves via the alias
- [x] Verified directly against the real DataForge cache's journal content: 26 RS lines produced, formatting confirmed correct for Agricium, Savrilium (alias), Quantainium, Ice, and Aphorite (correctly untagged)


Osiris-DevWorks (dismissed):
Reviewed as the diff on top of #238, per the PR description. This is a well-built extension:

- The table expansion is honestly sourced: 7 of the original 8 values cross-check exactly against the starminersdepot Alpha 4.8 reference, Ice's unverifiable value is kept rather than guessed at, and the 9 ores with no known value get nothing instead of a placeholder. That's the right call every time.
- The insertion point (right after the EM3 name header, before Locations) reads naturally in the block format, and the EM4 value matches the journal's existing accent styling.
- The Savrilium alias for CIG's own typo is a nice catch, and it's locked by a dedicated test.
- The MINEABLE_RS_VALUES docstring now names both consumers and points at each one's unmatched-key handling, and scripts/CLAUDE.md got its own section. Exactly the documentation shape this table needs now that two features read it.

Full suite is green on this branch on my machine (1116 passed, 16 skipped).

One wording nit, take it or leave it: the alias-map comment says the journal spells "a couple of ore names" differently, but the map has exactly one entry. Worth tightening to "one known mismatch" so nobody goes hunting for the second.

Merge order: this stacks on #238, which stacks on #232, so it merges after both and then rebases so the diff collapses to the compendium work. Approving on that basis.

PR #244, opened 2026-07-15
Fix: FPS armor "Core" pieces showing in Blueprint Tracker's Other bucket

Stealrull:
## Bug

A few FPS armor sets were showing up in the Blueprint Tracker's "Other" Type filter instead of "Armor".

## Root cause

Newer manufacturer-prefixed FPS armor sets (Antium, Testudo, Palatino, Aril, Geist, Overlord, and others) name their torso-platform piece "Core", and their loc keys carry no `"armor"` token at all:

```
item_Name_qrt_specialist_heavy_core_01_01_01  -> "Antium Core"
item_Name_kap_combat_light_core_01_01_01      -> "Geist Armor Core"
item_Name_GRIN_utility_medium_core_01_01_01   -> "Aril Core"
```

CIG's older armor lines *do* carry `"armor"` in the key (`item_Name_cds_armor_heavy_core_...`), so this was an inconsistency in CIG's own naming, not a universal gap. `blueprint_type_from_key` had no word covering `"core"`, so these pieces fell into `None` -> `"Other"`, while their Helmet/Backpack siblings from the exact same sets correctly showed as Armor (since `"helmet"` and `"backpack"` are already covered).

## Fix

Added `"core"` to `_ARMOR_EXTRA_WORDS` in `src/utils/blueprint_meta.py`.

Verified this is safe against ship components that also use "Core" in their display name (LuxCore, CoolCore, ThermalCore) — those carry a component-type-code prefix (`POWR_`/`COOL_`) that `component_type_from_key` already resolves *before* this word list is ever checked. Swept every `item_Name*` key in the real DataForge cache containing "core" (462 keys) — every single one is either an already-resolved ship component or an FPS armor torso piece; nothing else uses "core".

## Test plan
- [x] `pytest tests/` — 1066 passed, 16 skipped
- [x] New assertions in `test_blueprint_meta.py::test_blueprint_type_armor_core_pieces` covering multiple manufacturer prefixes, an older set that already had "armor" in the key (no regression), and both ship-component "Core" name collisions
- [x] Verified directly against the real DataForge cache: all previously-affected items (Antium Core, Testudo Core, Geist Armor Core, Palatino Core, Aril Core, TrueDef-Pro Core, Overlord Core Supernova) now resolve to "Armor"; ship power plant "LuxCore" still resolves to "Power Plant"

### DIFF
--- a/src/utils/blueprint_meta.py
+++ b/src/utils/blueprint_meta.py
@@ -32,8 +32,16 @@ from src.models.string_model import (
 # Extra armour-piece key tokens beyond string_model's set (which is tuned for
 # the strings-table category, not blueprint classification). Armour blueprints
 # are keyed by piece (backpack / undersuit / ...) with no "armor" token, so
-# without these they fell into the "Other" type bucket (#195).
-_ARMOR_EXTRA_WORDS = ("backpack", "undersuit", "flightsuit", "torso", "_legs", "_arms")
+# without these they fell into the "Other" type bucket (#195). "core" covers
+# the torso-platform piece of newer manufacturer-prefixed FPS armor sets
+# (e.g. item_Name_qrt_specialist_heavy_core_01_01_01 -> "Antium Core",
+# item_Name_kap_combat_light_core_01_01_01 -> "Geist Armor Core") whose keys
+# carry no "armor" token either — CIG's older sets do (cds_armor_heavy_core_*),
+# so this was inconsistent rather than universally missing. Safe against ship
+# components sharing the "core" word (LuxCore, CoolCore, ThermalCore) since
+# those carry a component-type-code prefix (POWR_/COOL_) resolved earlier by
+# component_type_from_key, which always wins first.
+_ARMOR_EXTRA_WORDS = ("backpack", "undersuit", "flightsuit", "torso", "_legs", "_arms", "core")
 from src.utils.owned_items import (
     BP_SECTION_HEADER,
     extract_bp_item_names,
--- a/src/utils/blueprint_meta.py
+++ b/src/utils/blueprint_meta.py
@@ -32,16 +32,23 @@ from src.models.string_model import (
 # Extra armour-piece key tokens beyond string_model's set (which is tuned for
 # the strings-table category, not blueprint classification). Armour blueprints
 # are keyed by piece (backpack / undersuit / ...) with no "armor" token, so
-# without these they fell into the "Other" type bucket (#195). "core" covers
-# the torso-platform piece of newer manufacturer-prefixed FPS armor sets
-# (e.g. item_Name_qrt_specialist_heavy_core_01_01_01 -> "Antium Core",
+# without these they fell into the "Other" type bucket (#195). "_core" (with
+# the leading underscore) covers the torso-platform piece of newer
+# manufacturer-prefixed FPS armor sets (e.g.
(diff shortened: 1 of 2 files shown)


src/utils/blueprint_meta.py
```diff
@@ -32,8 +32,16 @@
...
+# without these they fell into the "Other" type bucket (#195). "core" covers
+# the torso-platform piece of newer manufacturer-prefixed FPS armor sets
+# (e.g. item_Name_qrt_specialist_heavy_core_01_01_01 -> "Antium Core",
+# item_Name_kap_combat_light_core_01_01_01 -> "Geist Armor Core") whose keys
+# carry no "armor" token either — CIG's older sets do (cds_armor_heavy_core_*),
+# so this was inconsistent rather than universally missing. Safe against ship
+# components sharing the "core" word (LuxCore, CoolCore, ThermalCore) since
+# those carry a component-type-code prefix (POWR_/COOL_) resolved earlier by
+# component_type_from_key, which always wins first.
+_ARMOR_EXTRA_WORDS = ("backpack", "undersuit", "flightsuit", "torso", "_legs", "_arms", "core")
```
Osiris-DevWorks: These words match by plain substring (`w in kl`), and bare "core" has a wider blast radius than the rest of this list: "score" contains "core" (s + core), so any future key with a score/scoreboard token that reaches this check would misclassify as Armor. The list already solves this exact problem for "_legs"/"_arms" with the underscore prefix.

Every armor key this fix targets carries the underscore form (`_core_01_...`, including the older `cds_armor_heavy_core_*`), so `"_core"` matches everything "core" needs to match while ruling out the score-shaped collisions and any future CamelCase component name (SomethingCore) whose prefix code isn't in the component map yet. One character, strictly safer:

```python
_ARMOR_EXTRA_WORDS = ("backpack", "undersuit", "flightsuit", "torso", "_legs", "_arms", "_core")
```

Your sweep shows today's cache is clean either way, so this is future-proofing, not a live bug. But since the tests all use `_core_` keys anyway, they pass unchanged with the tighter token.

Osiris-DevWorks (dismissed):
Reviewed. Right diagnosis, right fix location, and the verification work is what sells it: the 462-key sweep of the real cache plus the LuxCore/CoolCore collision check answers the exact question I'd have asked. The test locks all three angles (new sets, the older armor-token sets, and the component-prefix collisions).

Full suite is green on this branch on my machine (1066 passed, 16 skipped).

One hardening suggestion, details inline. Approving either way since the current cache is verifiably safe; if you take the suggestion, do it before merge so the test lands with it.

src/utils/blueprint_meta.py
Stealrull: Completely agree, will add to futureproof not a hard thing.

Stealrull: Done in 633ad15 — `_ARMOR_EXTRA_WORDS` now has `"_core"` instead of bare `"core"`. Added an explicit test locking the score/highscore non-collision alongside the existing coverage. Full suite green (1066 passed, 16 skipped).

Osiris-DevWorks (approved):
Re-approving on the new head: the "_core" tightening is exactly the hardening I suggested, the comment now documents the score-collision reasoning, and the new negative tests lock it. Full suite green on this branch on my machine (1066 passed, 16 skipped). Ready to merge; #246 follows it.

PR #245, opened 2026-07-15
Fix: crossbow FPS weapons showing in Blueprint Tracker's Other bucket

Stealrull:
## Bug

New crossbow FPS weapons showing under "Other" in the Blueprint Tracker's Type dropdown instead of "FPS Weapon".

## Root cause

Same shape as #244 (armor "Core" pieces) — CIG's crossbow weapon (`utfl_crossbow_ballistic_01`, plus `_mag`/`_short`/`_sota01` skin variants) has no matching token in `_FPS_WEAPON_WORDS`:

```
item_Nameutfl_crossbow_ballistic_01        -> falls through to None -> "Other"
item_Nameutfl_crossbow_ballistic_01_mag
item_Nameutfl_crossbow_ballistic_01_short
item_Nameutfl_crossbow_ballistic_01_sota01
```

## Fix

Added `"_crossbow_"` to `_FPS_WEAPON_WORDS` in `src/models/string_model.py` (shared between the main table's category classifier and `blueprint_meta.py`'s Blueprint Tracker Type classifier).

The main String Editor table's Category column was already unaffected — its broader lowercase-fallback rule already returned "Gear" for these keys regardless of the specific word-list match. Only the Blueprint Tracker's more granular Type bucket (which has no such fallback) was actually broken.

## Test plan
- [x] `pytest tests/` — 1065 passed, 16 skipped
- [x] New assertion in `test_blueprint_meta.py::test_blueprint_type_buckets`
- [x] Verified directly against the real DataForge cache: all 4 crossbow item keys now resolve to "FPS Weapon"

### DIFF
--- a/src/models/string_model.py
+++ b/src/models/string_model.py
@@ -12,7 +12,7 @@ from functools import lru_cache
 
 _FPS_WEAPON_WORDS = (
     "_rifle_", "_pistol_", "_smg_", "_shotgun_", "_sniper_",
-    "_launcher_", "_lmg_", "_hmg_", "_knife_", "_multi_",
+    "_launcher_", "_lmg_", "_hmg_", "_knife_", "_multi_", "_crossbow_",
 )
 _ARMOR_GEAR_WORDS = (
     "armor", "helmet", "suit", "vest", "glasses", "_optics_", "_barrel_",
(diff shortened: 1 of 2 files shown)


Osiris-DevWorks (approved):
Reviewed. Clean, minimal, and it uses the underscore-delimited token form (`_crossbow_`) that keeps substring matching tight, which is exactly the pattern I want these word lists converging on (see my note on #244).

I verified the shared-list concern independently: `_FPS_WEAPON_WORDS` also drives the main table's Category classifier, but in string_model an FPS-word match returns "Gear", the same thing the fallback already returned for these keys, so the Category column is byte-identical before and after. Only the Blueprint Tracker's Type bucket changes, which is the bug. The PR description called this out unprompted, which I appreciate.

Full suite is green on this branch on my machine (1065 passed, 16 skipped). Approving.

PR #246, opened 2026-07-15
Fix: Blueprint Tracker false positives + Carnifex armor pieces in Other

Stealrull:
Stacks on #244 (fix/blueprint-armor-core-classification) since it extends the same `_ARMOR_EXTRA_WORDS` list — review as the diff on top of #244; #244's own commit already has its review there.

Two independent bugs reported together in the same screenshot.

## Bug 1: non-blueprint text leaking into the Blueprint Tracker

`extract_bp_item_names` / `apply_owned_to_value` matched every `\n- text` bullet **anywhere** in a mission body once a POTENTIAL BLUEPRINTS header existed, not just bullets inside that section. Two ways this leaked non-blueprint text in:

- A stray `- Stows` / `- Stow` / `- Stows out.` line in the mission's flavor-text prose, **before** the header (a CIG data quirk — looks like leftover fragment text).
- A real bullet from a **later** section (`- Council Scrip` / `- MG Scrip` under `ITEM REWARDS`), **after** the blueprints list.

### Fix

Scoped the bullet scan to the span between the `POTENTIAL BLUEPRINTS` header and the next genuine section header — distinguished from the per-region sub-headers that live *inside* the blueprints list itself (`<EM4>[Nyx]</EM4>`, `<EM4>[Pyro RegionA, Pyro RegionB]</EM4>`) by checking whether the tag's content starts with `"["` (sub-headers do, real section headers like `ITEM REWARDS`/`MISSION DETAILS` never do).

## Bug 2: Carnifex armor pieces in "Other"

"Carnifex Armor Core" / "Carnifex Armor Legs" (`item_Name_gys_jacket_01_01_01` / `item_Name_gys_pants_01_01_01`) fell into "Other" — yet another manufacturer-specific naming scheme (jacket/pants) with no "armor" token, same shape as #244's Core-piece fix.

### Fix

Added `"gys_jacket"` / `"gys_pants"` to `_ARMOR_EXTRA_WORDS`, scoped to the `"gys_"` prefix rather than bare `"jacket"`/`"pants"` — those bare words also match hundreds of unrelated civilian wardrobe items (987/ALB/CTL/DMC fashion lines) elsewhere in `item_Name` space. Confirmed none of those ever appear as real blueprint rewards today, but scoping to the manufacturer prefix removes the risk entirely rather than relying on that happening to stay true.

## Test plan
- [x] `pytest tests/` — 1070 passed, 16 skipped
- [x] New tests: `TestExtract`/`TestApply` false-positive scoping (prose bullet, later-section bullet, multi-region sub-headers still work) in `test_owned_items.py`; `test_blueprint_type_gys_jacket_and_pants` in `test_blueprint_meta.py`
- [x] Verified directly against the real DataForge cache: all 5 previously-leaking names (Council Scrip, MG Scrip, Stow, Stows, Stows out.) are gone from the extracted blueprint set; Carnifex Armor Core/Legs are still present and now correctly typed as "Armor"

### DIFF
--- a/src/utils/blueprint_meta.py
+++ b/src/utils/blueprint_meta.py
@@ -32,8 +32,16 @@ from src.models.string_model import (
 # Extra armour-piece key tokens beyond string_model's set (which is tuned for
 # the strings-table category, not blueprint classification). Armour blueprints
 # are keyed by piece (backpack / undersuit / ...) with no "armor" token, so
-# without these they fell into the "Other" type bucket (#195).
-_ARMOR_EXTRA_WORDS = ("backpack", "undersuit", "flightsuit", "torso", "_legs", "_arms")
+# without these they fell into the "Other" type bucket (#195). "core" covers
+# the torso-platform piece of newer manufacturer-prefixed FPS armor sets
+# (e.g. item_Name_qrt_specialist_heavy_core_01_01_01 -> "Antium Core",
+# item_Name_kap_combat_light_core_01_01_01 -> "Geist Armor Core") whose keys
+# carry no "armor" token either — CIG's older sets do (cds_armor_heavy_core_*),
+# so this was inconsistent rather than universally missing. Safe against ship
+# components sharing the "core" word (LuxCore, CoolCore, ThermalCore) since
+# those carry a component-type-code prefix (POWR_/COOL_) resolved earlier by
+# component_type_from_key, which always wins first.
+_ARMOR_EXTRA_WORDS = ("backpack", "undersuit", "flightsuit", "torso", "_legs", "_arms", "core")
 from src.utils.owned_items import (
     BP_SECTION_HEADER,
     extract_bp_item_names,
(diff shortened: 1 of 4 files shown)


Osiris-DevWorks (dismissed):
Reviewed as the diff on top of #244, per the description. Both fixes are right, and the section-scoping one is genuinely well designed:

- _bp_section_span bounds the bullet scan between the POTENTIAL BLUEPRINTS header and the next real section header, and the "content starts with [" test is the correct discriminator: it keeps the per-region sub-headers ([Nyx], [Pyro ...]) inside the span, and it also means an inline <EM4>[Owned]</EM4> tag on an already-tagged bullet can never be mistaken for a section boundary. I checked that interaction specifically; it holds on both the extract and apply paths.
- The apply-side reconstruction (prefix + retagged span + suffix) keeps retagging pure and idempotent, and the new test covering an owned set that happens to contain "Council Scrip" is exactly the nasty case worth locking.
- Scoping the Carnifex tokens to the gys_ manufacturer prefix instead of bare "jacket"/"pants" is the right instinct, for the same substring-collision reason as the _core tightening on #244. Glad to see that one adopted here too.
- The test fixture reproducing the real screenshot shape (prose bullet before the header, multi-region sub-headers, ITEM REWARDS bullet after) is the best kind of regression test.

Full suite is green on this branch on my machine (1070 passed, 16 skipped).

Merge order: this stacks on #244, so #244 goes first and this follows. Approving.

PR #253, opened 2026-07-16
[#251] Survive corrupt/non-UTF-8 bytes in base.ini and all state files

### LINKED ISSUE #251: Traceback error
*Synced from Discord: [Traceback error](https://discord.com/channels/1438175448420057323/1527310177970294947)*

User on facebook reported this error with 2.2, want to look into it.

Traceback (most recent call last):
  File "src\gui\workers.py", line 398, in run
  File "C:\Users\UserName\AppData\Local\Osiris DevWorks\Smart Citizen\_internal\scripts\generate_enhancements_ini.py", line 6520, in main
    loc = parse_ini(base_ini_path)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\UserName\AppData\Local\Osiris DevWorks\Smart Citizen\_internal\scripts\generate_enhancements_ini.py", line 155, in parse_ini
    for line in f:
  File "<frozen codecs>", line 322, in decode
  File "encodings\utf_8_sig.py", line 69, in _buffer_decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa0 in position 3517: invalid start byte

Stealrull:
## Summary
- Fixes the 2.2.0 blocker in #251: one corrupt byte (`0xA0`) in a user's cached `base.ini` crashed Generate Enhancements with `UnicodeDecodeError`. Investigation (comparing against a healthy install's extraction of the same LIVE data, which decodes as perfectly valid UTF-8) points to **machine-local corruption** — possibly in the user's own `Data.p4k`, so re-extraction can reintroduce the bad byte. The app must tolerate it.
- **Smart decode fallback, not a blanket one:** the two realistic failure shapes need opposite handling, discriminated by measuring whether the UTF-8 replacement count tracks the file's high-byte count:
  - *Corrupted UTF-8* (the actual report: one bad byte in 10.4 MB of valid UTF-8) → UTF-8 `errors="replace"`. Only the corrupt byte degrades to U+FFFD. A blanket cp1252 fallback here would mojibake every legitimate `é`/`™`/em-dash in the file (thousands of strings) to repair one byte.
  - *Genuinely ANSI file* (e.g. re-saved by legacy Notepad) → cp1252 decode, since there nearly every high byte is individually invalid UTF-8.
- **Audit of every other read of user-mutable/corruptible state found five more members of the same bug class, all fixed** — this was a full app blocker and each of these was an independent way to stay blocked:
  - `src/merger/ini_merger.py` — the same corrupt `base.ini` also failed **Apply to Game** (strict `open()` → `IOError`).
  - `src/utils/user_ini_manager.py` — `_backup_if_changing` runs on **every save**; a corrupt `user.ini` raised `UnicodeDecodeError` past its `OSError`-only guard, bricking all saves. Now compares newline-normalized bytes (no decode at all); a corrupt file counts as "changing" and is snapshotted before the clean rewrite.
  - `src/utils/json_settings.py` — a corrupt `settings.json` escaped the `(OSError, JSONDecodeError)` guard and **crashed the portable app at startup**.
  - `src/utils/dataforge_diff.py` — a corrupt diff manifest crashed the enhancements run; now treated as "no manifest" (regenerate everything; `update_manifest` rewrites a clean one — self-healing).
  - Stamp/patch reads in `settings.py`, the generator, and `dataforge_patcher.py` — `OSError`/`JSONDecodeError`-only guards widened to `ValueError`.
- The shared helper lives in `src/utils/ini_io.read_ini_text` (parser + merger both use it; merger already imports `src.utils`, and the parser→merger import direction forbids the reverse). The generator script keeps its own documented mirror to stay stdlib+lxml-only for standalone runs — same tolerated-duplication shape as `CATEGORY_SUBTREES` ↔ `DATAFORGE_KEEP_SUBPATHS`.
- Paths audited and already safe: `blueprint_log_scanner` (`errors="replace"`), `user_cfg` (broad except; skipping is safer than rewriting a mangled decode into the game's own config), i18n/tutorial/docs loads, `pak_extractor` stamps, `applied_file_validator`, `should_autosave_user_ini` (stat only). Registry-mode settings have no file decode to corrupt, so both build modes are covered.

**Release-target note:** opened against `release/2.3.0` per the branching model. This is a self-contained bug fix affecting current 2.2.0 users (full blocker for the affected user), but with `release/2.2.1` retired, all fixes are consolidating onto 2.3.0 rather than a separate hotfix branch.

Fixes #251.

## Testing Checklist
- [x] PR is scoped to one concern (one bug class: strict decodes of user-mutable state)
- [x] `pytest tests/` passes locally (1204 passed, 16 skipped)
- [ ] Affected user-facing flow exercised manually (Generate Enhancements / Apply with a deliberately corrupted base.ini)
- [x] User-facing strings, `docs/HELP.md`, `docs/ABOUT.md`, and `src/gui/coach_mark.py` reviewed for drift (no UI strings changed)
- [x] New non-exempt code has matching test coverage in `tests/` (17 tests in `test_ini_encoding_fallback.py` covering both parsers, the Apply-path merger, portable settings, the diff manifest, stamps, and the save-path backup guard)
- [x] Target branch is the active `release/2.3.0`, not `main` (see release-target note above)

## Testing performed
Full suite (1204 passed, 16 skipped). The new tests reproduce the exact #251 file shape (one bare `0xA0` in otherwise-valid UTF-8 with multi-byte characters that must survive un-mojibake'd, and keys after the bad byte that a truncating parser would lose), plus the ANSI-file shape, BOM hybrids, cp1252's undefined bytes, the Apply-path merger read, portable-mode settings corruption, corrupt diff-manifest self-healing, corrupt-stamp reads, and the every-save backup guard (including that identical re-saves still don't churn the backup rotation). Not exercised: a live GUI run against a real corrupted install.

## Post-merge QA
- [ ] With a deliberately corrupted `base.ini` (flip one byte ≥ 0x80): app loads all strings, Generate Enhancements completes, Apply to Game completes; Log tab shows the corruption warning
- [ ] Portable build: corrupt `settings.json` by hand → app starts with default settings instead of crashing
- [ ] For the original reporter: RSI Launcher **Verify** remains the real cure if their `Data.p4k` is corrupt — the app now degrades one string instead of dying, but their game files are still damaged

---

### DIFF
--- a/scripts/generate_enhancements_ini.py
+++ b/scripts/generate_enhancements_ini.py
@@ -149,19 +149,45 @@ OUTPUT_DIR = APP_CACHE_DIR
 
 # ── INI helpers ───────────────────────────────────────────────────────────────
 
+def _read_ini_text(path: Path) -> str:
+    """Read an INI file's text, tolerating non-UTF-8 content (#251).
+
+    base.ini is normally UTF-8 (with BOM) straight out of Data.p4k, but the
+    cached copy can stop being valid UTF-8 on a user's machine — re-saved by
+    an external editor in ANSI (legacy Notepad's default), mangled by a
+    sync/AV tool, or a community-translated base saved as ANSI. CIG's text
+    is full of characters that cp1252 encodes as single high bytes (é / ™ /
+    ° / em dash / non-breaking space — the reported crash was byte 0xA0, an
+    NBSP, on a stock English install), and any one of them is an invalid
+    UTF-8 start byte, so a strict decode crashed the whole enhancements
+    run. Fall back to cp1252 (errors="replace" guards its five undefined
+    bytes) rather than dying on the first non-UTF-8 byte.
+    """
+    raw = path.read_bytes()
+    try:
+        return raw.decode("utf-8-sig")
+    except UnicodeDecodeError as e:
+        logger.warning(f"{path} is not valid UTF-8 ({e}); decoding as Windows-1252")
+        if raw.startswith(b"\xef\xbb\xbf"):
+            raw = raw[3:]
+        return raw.decode("cp1252", errors="replace")
+
+
 def parse_ini(path: Path) -> dict[str, str]:
     result = {}
-    with open(path, encoding="utf-8-sig") as f:
-        for line in f:
-            line = line.rstrip("\r\n")
-            if not line.strip() or line.strip().startswith(";"):
-                continue
-            if "=" not in line:
-                continue
-            k, _, v = line.partition("=")
-            lookup_key = k.strip().split(",")[0].strip()
-            if lookup_key:
-                result[lookup_key] = v.strip()
+    # split('\n') + rstrip('\r'), NOT str.splitlines(): splitlines also
+    # breaks on U+2028/U+0085 etc., which a loc value could legitimately
+    # contain — file iteration never split on those and neither do we.
+    for line in _read_ini_text(path).split("\n"):
+        line = line.rstrip("\r")
+        if not line.strip() or line.strip().startswith(";"):
+            continue
+        if "=" not in line:
+            continue
+        k, _, v = line.partition("=")
+        lookup_key = k.strip().split(",")[0].strip()
+        if lookup_key:
+            result[lookup_key] = v.strip()
     return result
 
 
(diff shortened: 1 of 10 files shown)


Stealrull:
Closing as superseded: `release/2.2.1` already carries this exact fix directly (commits `4f70651`/`da8cd4f`, version bump `eb26a94`), landed independently of this fork's branch. The one thing from this branch not yet upstream — the `dataforge_patcher` corrupt-patch-file test found during the pre-release test-coverage audit — has been carried forward into #279 against `release/2.2.1`.

PR #252, opened 2026-07-16
[#247] i18n: sweep remaining hardcoded UI strings that bypass tr()

Stealrull:
## Summary
- The issue named 5 known offender groups in `src/gui/` bypassing `tr()` (Import INI flow, the generic error dialog title, 5 status-bar messages, 3 Tag Builder tooltips, the Success backup-restored box). A full search found far more: **~440 distinct `tr()` call sites now cover literals across `main_window.py`, `enhancements_tab.py`, `config_tab.py`, and `test_plan_panel.py`** (the last of which had zero i18n before this PR).
- **Scope note (per CONTRIBUTOR_GUIDE's size-signal guidance):** this is intentionally one wide-blast PR rather than several small ones — the sweep only makes sense evaluated as a whole (one file's fix often revealed the same gap pattern in a sibling file), and splitting it would create more overlapping-diff review pain than one large, mechanical PR. 6 files changed, ~1010 net lines.
- **Two real bugs fixed along the way**, not just string wrapping:
  1. `main_window.py`: the "Star Citizen Path Required" dialog switched to the Config tab by comparing tab text against the literal `"Config"` — silently breaks once that text is translated. Now uses `self.tabs.indexOf(self.config_tab)`, matching the pattern already used for the Log tab elsewhere in the file.
  2. `main_window.py`: the Test Plan dock's title was a bare `QDockWidget("Test Plan", ...)` literal with no retranslate coverage at all (same root cause as the rest of this issue).
- **Retranslate cascade extended**, not just literals wrapped: converting a hardcoded string to `tr()` only half-fixes the bug if the widget is never revisited on a runtime language switch — it'll show the *first* language forever until restart, arguably worse than the prior all-English behavior. Extended `retranslate_ui()` cascades in `main_window.py`, `enhancements_tab.py` (including new `retranslate_ui()` methods on `_TagBuilderPage` and `_ElementRow`, which had none before — there was nothing to retranslate), and `test_plan_panel.py` (new, from scratch).
- **Explicitly out of scope** (flagged, not silently pulled in — each is a separate, larger conversion):
  - `src/utils/tag_builder.py`'s option tables (`SEPARATORS`, `ENCLOSINGS`, `PLACEMENTS`, `ROUTE_ARROWS`, `REMOVE_WORD_OPTIONS`, etc.) back several Tag Builder combo boxes but live in a domain-constants layer, not `src/gui/`.
[...]

PR #260, opened 2026-07-17
[#259] Fix Mission Titles page: 7 checkboxes never armed Save Tag Changes

### LINKED ISSUE #259: Mission Titles page: several checkboxes don't mark Save Tag Changes dirty
## What

On the Enhancements tab's Mission Titles page, toggling any of these checkboxes does **not** light up the "Save Tag Changes" button (it stays disabled/grey, giving no indication there's anything to save):

- **Underline "Direct"** (reported)
- Shorten original titles
- Shorten cargo sizes
- The three "remove word" checkboxes (Cargo / Haul / Rank)

## Root cause

`src/gui/enhancements_tab.py`: every other control on this page (the 5 dropdowns, the General Tags / Scanning checkboxes) calls `self.config_changed.emit()` after updating `self.config`, and `config_changed` is what MainWindow listens to via `_mark_tag_dirty()` to enable the Save Tag Changes button. Three handler methods update `self.config` and refresh the preview but never emit the signal:

- `_on_mt_abbrev_toggle()` — used by the Cargo/Haul/Rank checkboxes *and* Underline Direct
- `_on_mt_shorten_titles_toggle()`
- `_on_mt_shorten_sizes_toggle()`

## Impact

Since `_apply_tag_btn.setEnabled(dirty)` disables the button entirely when not dirty, a user who *only* toggles one of these six checkboxes has no way to save the change via Save Tag Changes — the button stays disabled. The change isn't lost forever (`_on_generate_enhancements_clicked` unconditionally persists the current in-memory config regardless of dirty state, so clicking **Generate Enhancements** does save it), but the dirty indicator itself is simply wrong for these six controls, and a user who expects Save Tag Changes to work without a full regen gets silently ignored.

Pre-existing bug — confirmed present on `main` before this cycle's other fixes.

Stealrull:
## Summary
- Reported: toggling **Underline "Direct"** on the Enhancements tab's Mission Titles page doesn't light up the **Save Tag Changes** button.
- Root cause: every other control on this page (the 5 dropdowns, General Tags / Scanning checkboxes) calls `self.config_changed.emit()` after updating `self.config`; MainWindow wires that signal to `_mark_tag_dirty()`, which is what enables the Save Tag Changes button (`_apply_tag_btn.setEnabled(dirty)` — it's flat-out **disabled** while not dirty). Four handlers update `self.config` and refresh the preview but never emitted the signal:
  - `_on_mt_standardize_toggle` (Standardize hauling mission names)
  - `_on_mt_abbrev_toggle` (backs Cargo / Haul / Rank *and* Underline Direct)
  - `_on_mt_shorten_titles_toggle` (Shorten original titles)
  - `_on_mt_shorten_sizes_toggle` (Shorten cargo sizes)
  
  **7 checkboxes total** silently failed to arm the dirty state — found the 4th handler (`_on_mt_standardize_toggle`) while fixing the reported one.
- Impact: the edit isn't lost forever — **Generate Enhancements** unconditionally persists the current in-memory config regardless of dirty state — but a user who only touches one of these 7 controls and expects **Save Tag Changes** to work without a full regen gets silently ignored, with the button giving no indication anything needs saving.
- Pre-existing bug, confirmed present identically on `main`, `release/2.2.1`, and `release/2.3.0` — predates this cycle's other work entirely.
- Scanned every other `_on_*` handler in the file for the same missing-emit pattern; none found.

Fixes #259.

## Testing Checklist
- [x] PR is scoped to one concern (one root cause: 4 handlers missing a signal emit)
- [x] `pytest tests/` passes locally (1213 passed, 16 skipped)
- [ ] Affected user-facing flow exercised manually (toggle Underline Direct on the Mission Titles page, confirm Save Tag Changes lights up)
- [x] User-facing strings, `docs/HELP.md`, `docs/ABOUT.md`, and `src/gui/coach_mark.py` reviewed for drift (none — no UI strings changed, purely a signal-wiring fix)
- [x] New non-exempt code has matching test coverage — added `tests/test_mission_titles_page_dirty_signal.py`, driving the real `_TagBuilderPage` widget headlessly (`QT_QPA_PLATFORM=offscreen`, same pattern as `tests/test_restore_backup.py`, no pytest-qt dependency needed)
- [x] Target branch is `release/2.2.1` (patch — bug fix only)

## Testing performed
Full suite (1213 passed, 16 skipped). Confirmed the 9 new tests fail against the pre-fix code (7 of them, exactly matching the 7 broken checkboxes) and pass after the fix, with the 2 dropdown/config-value sanity checks passing throughout — ruling out both a false-positive test and an accidental regression to the controls that already worked. Not exercised: a live GUI toggle-and-observe-the-button-light-up pass (no build environment here).

## Post-merge QA
- [ ] On the Mission Titles page, toggle each of: Underline Direct, Shorten original titles, Shorten cargo sizes, Standardize hauling mission names, and the Cargo/Haul/Rank checkboxes — confirm Save Tag Changes turns red/enabled after each
- [ ] Confirm the other Tag Builder pages (components, missiles, ship weapons, commodities) and their dropdowns still correctly arm Save Tag Changes — unaffected by this change, but worth a quick regression pass since they share the same button

---

### DIFF
--- a/src/gui/enhancements_tab.py
+++ b/src/gui/enhancements_tab.py
@@ -1552,6 +1552,7 @@ class _TagBuilderPage(QWidget):
     def _on_mt_standardize_toggle(self, checked: bool) -> None:
         self.config.standardize_hauling_names = checked
         self._refresh_preview()
+        self.config_changed.emit()
 
     def _on_mt_abbrev_toggle(self, key: str, checked: bool) -> None:
         phrases = set(getattr(self.config, "abbreviated_phrases", frozenset()))
@@ -1561,6 +1562,7 @@ class _TagBuilderPage(QWidget):
             phrases.discard(key)
         self.config.abbreviated_phrases = frozenset(phrases)
         self._refresh_preview()
+        self.config_changed.emit()
 
     def _on_mt_shorten_titles_toggle(self, checked: bool) -> None:
         phrases = set(getattr(self.config, "abbreviated_phrases", frozenset()))
@@ -1568,10 +1570,12 @@ class _TagBuilderPage(QWidget):
         phrases = (phrases | keys) if checked else (phrases - keys)
         self.config.abbreviated_phrases = frozenset(phrases)
         self._refresh_preview()
+        self.config_changed.emit()
 
     def _on_mt_shorten_sizes_toggle(self, checked: bool) -> None:
         self.config.shortened_sizes = frozenset(_ALL_SIZE_WORDS) if checked else frozenset()
         self._refresh_preview()
+        self.config_changed.emit()
 
     def _on_mt_rank_sep_changed(self, _idx: int) -> None:
         self.config.rank_separator = self._mt_rank_sep.currentData() or self.config.rank_separator
--- a/tests/test_mission_titles_page_dirty_signal.py
+++ b/tests/test_mission_titles_page_dirty_signal.py
@@ -1,0 +1,94 @@
+"""Mission Titles page: checkbox toggles must mark Save Tag Changes dirty (#259).
+
+Reported: toggling "Underline Direct" on the Enhancements tab's Mission
+Titles page didn't light up the Save Tag Changes button. Every other
+control on this page (the 5 dropdowns, General Tags / Scanning checkboxes)
+emits ``config_changed`` after updating ``self.config``, and MainWindow
+wires that signal to ``_mark_tag_dirty()`` (which enables the button — it's
+disabled while not dirty). Four handler methods updated ``self.config`` and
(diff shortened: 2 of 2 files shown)


Stealrull:
Superseded — this fix is already included in #252 (`issue/247-i18n-hardcoded-string-sweep` → `release/2.3.0`), reapplied directly onto that branch since it touches the same handler bodies the i18n sweep translated. `release/2.2.1` is being retired in favor of consolidating everything onto `release/2.3.0`. Closing this one — no separate PR needed, #252 already covers it.

PR #262, opened 2026-07-17
[#261] Write applied global.ini with a UTF-8 BOM — fixes @KeyName corruption

### LINKED ISSUE #261: Critical: applied global.ini missing UTF-8 BOM — game shows every string as raw @KeyName
## What

Confirmed on a tester's real Star Citizen install (not yet published to end users): after Apply, the in-game main menu, mobiGlas, contacts panel — effectively every piece of UI text — showed its raw loc key instead of resolved text (e.g. \`@Frontend_PU_SubHeader_Desc\`, \`@FRONTEND_JOIN_THE_UNIVERSE\`, \`@FRONTEND_PLAY_STAR_CITIZEN\`). Removing Smart Citizen's \`global.ini\` (reverting to vanilla) fixed it immediately.

## Root cause

Data.p4k's own extracted \`global.ini\` ships with a **UTF-8 BOM** (confirmed directly: raw unp4k output starts with \`\xef\xbb\xbf\`). Star Citizen's own localization loader appears to require that BOM to reliably detect the file's encoding — without it, the engine can fail to resolve the entire loc table (every string falls back to its raw \`@KeyName\` placeholder) rather than degrading per-key.

\`merge_ini_files\` in \`src/merger/ini_merger.py\` — the function that writes the final applied \`global.ini\` to the game's install folder — has always written plain UTF-8 with **no BOM** (\`open(output_path, 'w', encoding='utf-8')\`), in every prior release, not just this cycle's changes.

This went undetected because our own safety net, \`validate_applied_file\`, only checks that every stock key is *present* in the written file — and our own readers (\`parse_ini_file\`/\`read_ini_text\`) are BOM-tolerant (\`utf-8-sig\`-aware), so a BOM-less file parses back with all keys intact and passes validation cleanly. The game's own parser is apparently stricter than ours, so validation reported success on a file the game itself couldn't properly load.

## Reproduction

Ran the exact Extract→Apply pipeline against the tester's real cached \`base.ini\`/enhancement files/\`user.ini\`: the pre-fix code produces a byte-for-byte no-BOM file; the fixed code produces a file starting with \`\xef\xbb\xbf\`, matching Data.p4k's own extraction format exactly, and \`validate_applied_file\` (now BOM-aware) reports PASS only when the BOM is present.

## Fix

- \`merge_ini_files\` now writes with \`encoding='utf-8-sig'\` (BOM included), matching CIG's own reference format.
- \`validate_applied_file\` now checks for the BOM as a first-class validation condition (not just key presence) — a missing BOM fails validation and triggers the existing rollback-to-backup path, so if this write path ever regresses again (a future refactor, a hand-edited file, an external tool overwriting the output), it fails loudly and safely instead of silently installing a file the game can't read.

Targeting \`release/2.2.1\` given severity.

Stealrull:
## Summary
- **Critical bug**, confirmed on a tester's real Star Citizen install (caught during internal testing of the v2.2.1 portable build — not published to end users): after Apply, the entire in-game UI showed raw loc keys instead of resolved text (\`@Frontend_PU_SubHeader_Desc\`, \`@FRONTEND_JOIN_THE_UNIVERSE\`, \`@FRONTEND_PLAY_STAR_CITIZEN\`, etc. — every visible string). Deleting Smart Citizen's \`global.ini\` restored correct text immediately.
- **Root cause**: Data.p4k's own extracted \`global.ini\` ships with a UTF-8 BOM (confirmed directly: raw unp4k output starts with \`\xef\xbb\xbf\`). Star Citizen's own localization loader appears to require that BOM to reliably detect the file's encoding — without it, the engine can fail to resolve the **entire** loc table rather than degrading per-key, which is exactly the observed symptom (every string, not just changed ones). \`merge_ini_files\` — the function that writes the final applied \`global.ini\` — has written plain UTF-8 with **no BOM** in every prior release, not something introduced this cycle.
- **Why nothing caught it**: the existing rollback-on-failure safety net, \`validate_applied_file\`, only checked stock-key *presence*. Our own readers (\`parse_ini_file\`/\`read_ini_text\`) are \`utf-8-sig\`-aware, so a BOM-less file parses back with every key intact and passes validation cleanly — the game's own parser is apparently stricter than ours, so validation reported success on a file the game couldn't actually load.
- **Fix, two parts**:
  1. \`merge_ini_files\` now writes with \`encoding='utf-8-sig'\`, matching Data.p4k's own format byte-for-byte.
  2. \`validate_applied_file\` now treats a missing BOM as its own validation failure (independent of key presence), triggering the existing rollback-to-backup path. If this write path ever regresses again, it fails loudly instead of silently installing a file the game can't read.
- Verified by replaying the exact Extract→Apply pipeline against the reporting tester's real cached \`base.ini\`, enhancement files, and \`user.ini\`: pre-fix code reproduces a byte-for-byte no-BOM output; post-fix code produces a correctly-BOM'd file that \`validate_applied_file\` confirms as PASS.

Fixes #261.

## Testing Checklist
- [x] PR is scoped to one concern (one root cause, two-part fix + hardened safety net)
- [x] \`pytest tests/\` passes locally (1210 passed, 16 skipped)
- [ ] App launches: \`python src/main.py\`
- [ ] Affected user-facing flow exercised manually (Apply to Game, then check in-game that the main menu/mobiGlas render normal text, not raw @KeyName)
- [x] User-facing strings, \`docs/HELP.md\`, \`docs/ABOUT.md\`, and \`src/gui/coach_mark.py\` reviewed for drift (none — no UI strings changed)
- [x] New non-exempt code has matching test coverage — 2 new tests in \`tests/test_ini_encoding_fallback.py\` (BOM is written; BOM is written even when the source lacks one) and 4 new tests in \`tests/test_applied_file_validator.py\` (missing BOM fails validation even with a perfect key match; present BOM + perfect match passes; BOM failure reports alongside missing-key failures; a read error during the BOM check doesn't crash). Updated all pre-existing validator tests to write their "valid" fixture files with the BOM, since that's now part of what "valid" means.
- [x] Target branch is \`release/2.2.1\` (patch — bug fix only)
- [x] No direct \`QSettings\` calls, no hard-coded column indices, no \`QProgressBar.setValue()\` from workers
- [x] \`VERSION.TXT\` matches the active release branch

## Testing performed
Full suite (1210 passed, 16 skipped). Reproduced the actual bug and its fix end-to-end using a tester's real files (not synthetic fixtures): pulled their cached \`base.ini\`, all 9 enhancement INIs, and \`user.ini\` from their portable install's data folder, replayed the exact merge pipeline \`apply_to_game\` runs, and confirmed (a) the pre-fix code path produces a byte-for-byte BOM-less file matching what's on disk after their actual Apply, and (b) the fixed code produces output starting with \`\xef\xbb\xbf\` that \`validate_applied_file\` now reports as a clean PASS. Not exercised: an actual live-game screenshot confirming the fixed build resolves strings correctly (no build/deploy environment here) — that's the one item on the post-merge checklist below.

## Post-merge QA
- [ ] Apply to a real Star Citizen install, launch the game, confirm the main menu/mobiGlas/contacts panel show real text — this is the one thing that must be re-verified in a live game session, since it's the whole point of the fix
- [ ] Confirm Export Loc-Pack (which reads the already-applied file directly) also produces a BOM'd zip — should be automatic since it doesn't re-write the file, but worth a quick sanity check
- [ ] Spot-check that re-applying over an already-BOM'd file (the normal "user clicks Apply again" case) still validates cleanly

---

### DIFF
--- a/src/merger/ini_merger.py
+++ b/src/merger/ini_merger.py
@@ -215,8 +215,9 @@ def merge_ini_files(
     """Merge source INI with overrides, preserving all lines.
 
     Reads source file line-by-line, replaces values for matching keys,
-    and writes to output as UTF-8. Strips comma-based metadata suffixes
-    (e.g., "key,P") from keys to match normalized override keys.
+    and writes to output as UTF-8 **with a BOM** (#261). Strips comma-based
+    metadata suffixes (e.g., "key,P") from keys to match normalized
+    override keys.
 
     Note: Variant key syncing happens in merge_sources_by_hierarchy(), so
     the overrides_dict already has synced values when this is called.
@@ -253,7 +254,21 @@ def merge_ini_files(
         else:
             trailing_newline = False
 
-        with open(output_path, 'w', encoding='utf-8') as outfile:
+        # utf-8-sig, NOT utf-8 (#261): Data.p4k's own extracted global.ini
+        # ships with a UTF-8 BOM (confirmed: raw unp4k output starts with
+        # \xef\xbb\xbf), and Star Citizen's own loc-string loader appears to
+        # need that BOM to reliably detect the file's encoding. Every prior
+        # release wrote plain utf-8 (no BOM) here — Python's own readers
+        # (read_ini_text / parse_ini_file, both utf-8-sig-aware) never
+        # noticed the mismatch, so validate_applied_file's key-presence
+        # check always reported success even when the applied file was
+        # missing the BOM CIG's engine relies on. Symptom when it bites: the
+        # game shows every string as its raw @KeyName placeholder — not
+        # garbled individual values, because the failure is at the
+        # whole-file encoding-detection stage, before any per-key lookup
+        # happens. See validate_applied_file's BOM check for the second half
+        # of this fix — a safety net in case this write path regresses again.
+        with open(output_path, 'w', encoding='utf-8-sig') as outfile:
             for i, line_rstrip in enumerate(source_lines):
                 # Match the old per-line shape: every line ended with '\n'
                 # except a final line with no trailing newline.
--- a/src/utils/applied_file_validator.py
+++ b/src/utils/applied_file_validator.py
@@ -11,6 +11,21 @@ from src.parser.ini_parser import parse_ini_file
 
 logger = logging.getLogger(__name__)
 
+# Data.p4k's own extracted global.ini ships with this UTF-8 BOM, and Star
+# Citizen's own loc-string loader appears to need it to reliably detect the
+# file's encoding — without it, the game can fail to resolve ANY key (every
+# string shows its raw @KeyName placeholder) rather than degrading per-key
+# (#261). merge_ini_files writes with utf-8-sig specifically to include this,
+# but our OWN readers (parse_ini_file/read_ini_text) are utf-8-sig-aware and
+# silently accept a file with the BOM stripped — so the key-presence check
+# below would report a BOM-less file as fully valid even though the game
+# can't parse it. This is a second, independent line of defense: if the BOM
+# is ever dropped again (a future refactor, a hand-edited file, an external
+# tool overwriting the output), this check fails loudly and the caller rolls
+# back to the last known-good backup instead of leaving a file installed
+# that Python can read fine but the game engine cannot.
+_UTF8_BOM = b"\xef\xbb\xbf"
+
 
 def validate_applied_file(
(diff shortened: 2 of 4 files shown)


Stealrull:
Superseded by #263, which already targets `release/2.3.0`. `release/2.2.1` is being retired — all fixes are consolidating onto `release/2.3.0`. Closing this one.

PR #263, opened 2026-07-17
[#261] Write applied global.ini with a UTF-8 BOM — fixes @KeyName corruption

### LINKED ISSUE #261: Critical: applied global.ini missing UTF-8 BOM — game shows every string as raw @KeyName

Stealrull:
## Summary
Same fix as [PR #262](https://github.com/Osiris-DevWorks/smart-citizen/pull/262) (closed unmerged), reapplied here since this branch predates the #251 encoding rewrite and still had the older `merge_ini_files` shape (plain `open()` calls, not `read_ini_text`). See #261 for the full root-cause writeup — short version below.

- **Critical bug**, confirmed on a tester's real Star Citizen install: after Apply, the entire in-game UI showed raw loc keys instead of resolved text (`@Frontend_PU_SubHeader_Desc`, `@FRONTEND_JOIN_THE_UNIVERSE`, etc.). Deleting Smart Citizen's `global.ini` restored correct text immediately.
- **Root cause**: Data.p4k's own extracted `global.ini` ships with a UTF-8 BOM. Star Citizen's own localization loader appears to require it to reliably detect the file's encoding — without it, the engine can fail to resolve the entire loc table rather than degrading per-key. `merge_ini_files` has written plain UTF-8 with no BOM in every prior release.
- This branch has since been merged up to the current `release/2.2.1` tip (bringing in #251's `read_ini_text` rewrite) and reconciled: reads now go through `read_ini_text` (tolerates non-UTF-8 content and strips a source BOM via its own `utf-8-sig` decode), and the output is written with `utf-8-sig` to add the BOM the game needs.
- `validate_applied_file` now treats a missing BOM as a first-class validation failure (independent of key presence), triggering the existing rollback-to-backup path.

Fixes #261.

## Testing Checklist
- [x] PR is scoped to one concern
- [x] `pytest tests/` passes locally (1213 passed, 16 skipped)
- [ ] Affected user-facing flow exercised manually (Apply to Game, then check in-game)
- [x] User-facing strings/docs reviewed for drift (none — no UI strings changed)
- [x] New non-exempt code has matching test coverage — new tests in `tests/test_ini_merger.py` (BOM written; BOM written even when source lacks one; source BOM doesn't leak into first key name; values/structure preserved alongside the BOM) plus `tests/test_ini_encoding_fallback.py` (trailing-newline shape preserved under the BOM-aware read, and an explicit assertion the output starts with the BOM bytes) and `tests/test_applied_file_validator.py` (same as #262)
- [x] Target branch is `release/2.2.1`

## Testing performed
Full suite (1213 passed, 16 skipped).

## Post-merge QA
- [ ] App launches from a clean checkout
- [ ] Apply to a real Star Citizen install, launch the game, confirm the main menu/mobiGlas/contacts panel show real text

---


Osiris-DevWorks (changes requested):
Thanks for taking this one on. You found the real cause and the fix works, with good tests. Two things to fix, two suggestions.

**Fix: VERSION.TXT should say 2.2.1, not 2.3.0.** The installer and update checker both read this file, so a wrong number breaks releases. Easiest is to remove the file from this PR.

**Fix: update your branch.** The release branch changed since you started (the #251 fix landed), so GitHub shows a conflict. After updating, the one change left in `ini_merger.py` is `'utf-8'` to `'utf-8-sig'` on the output file. Keep your new tests.

**Suggestion:** the safety check passes when `base.ini` is missing or unreadable, even if the BOM is absent. It could report the missing BOM in those cases.

**Suggestion:** two tests have the BOM typed as an invisible character inside a quoted string. Writing it as `b"\xef\xbb\xbf"` makes it visible so it can't get lost.

Once those two fixes are in, this closes #261.

Stealrull:
Addressed all four items:
- VERSION.TXT now reads 2.2.1
- Branch updated to current release/2.2.1 (merged in, resolved the ini_merger.py overlap by keeping the utf-8-sig write plus read_ini_text's tolerant read)
- BOM safety check now reports the missing BOM even when base.ini is missing/unreadable (previously returned early with no warning)
- Both tests now write the BOM as explicit b"\xef\xbb\xbf" bytes / chr(0xFEFF) instead of an invisible character in the source

Osiris-DevWorks (approved):
All four review items addressed, and the validator rework came back stronger than asked: the BOM warning now survives every fail-open path, each locked by a test. Approving. This PR and #264 both append test classes to test_ini_merger.py; I'll resolve that overlap at merge time.

Issue #267, opened 2026-07-17
Xenothreat blueprints not showing


Stealrull:
Some of these show if doing a owned blueprint scan but should show in unowned otherwise, had a user on facebook bring this up.

Stealrull:
Confirmed fixed in 2.2.1 build

PR #271, opened 2026-07-17
[#269] fix: auto-detect Star Citizen install across all local drives

### LINKED ISSUE #269: Auto-detect Star Citizen install path across all drives
Found while testing a portable build: a tester's real Star Citizen install lives at \`E:\Games\Roberts Space Industries\StarCitizen\` — a secondary drive, nested under a personal "Games" folder. Smart Citizen's only auto-detection fallback checked two hardcoded \`C:\Program Files\...\` paths, so a fresh profile (or a portable build's first run, with no saved settings yet) showed the "Star Citizen Path Required" dialog even though a real install existed right there.

Fix: scan every local drive letter for a handful of common install shapes (RSI Launcher's own default under Program Files, a secondary drive kept in the same shape, and one nested under "Games"), validated the same way every other candidate already is. Persists the result once found so the scan only runs once per profile.

Stealrull:
## Summary
- The only auto-detection fallback checked two hardcoded `C:\Program Files\...` paths, so a fresh profile (or a portable build's first run, with no saved settings yet) showed the "Star Citizen Path Required" dialog even when a real install existed — found while testing #267's build: a tester's real install lives at `E:\Games\Roberts Space Industries\StarCitizen`, a shape neither hardcoded candidate could ever match.
- `_scan_common_sc_install_locations()` now checks every local drive letter against RSI Launcher's own default shape (both Program Files bitness variants), a bare secondary-drive install, and one nested under a personal "Games" folder — validated the same way every other candidate already is (`_is_valid_sc_root`).
- Wired into both `get_sc_install_root()` and `get_game_install_path()`, persisting the result once found so the scan only ever runs once per profile.

Closes #269.

## Testing Checklist
- [x] PR is scoped to one concern (single bug fix)
- [x] `pytest tests/` passes locally (1210 passed, 16 skipped on this branch)
- [x] Affected user-facing flow exercised manually: confirmed by the reporting user on a real portable build — install path now auto-detected with zero configuration
- [x] User-facing strings, `docs/HELP.md`, `docs/ABOUT.md`, and `src/gui/coach_mark.py` reviewed for drift (no UI strings/controls changed)
- [x] New non-exempt code has matching test coverage in `tests/` (new `TestScanCommonScInstallLocations` + `TestGetScInstallRootUsesDriveScan` classes in `test_sc_install_root.py`; 3 pre-existing tests updated since their `Path.exists`-faking approach no longer fully blocks auto-detection now that `_is_valid_sc_root` — which uses `is_dir()`, not `exists()` — drives the scan)
- [x] Target branch is the active `release/2.2.1`, not `main`

## Testing performed
Full suite (1210 passed, 16 skipped). Built a portable test build and the reporting user confirmed the fix on their real install (a secondary-drive install under a "Games" folder) — the app now finds it automatically with no manual configuration.

## Post-merge QA
- [ ] Fresh profile / portable first run with a real SC install on a non-C: drive: confirm no "Star Citizen Path Required" dialog appears and Extract works immediately
- [ ] Confirm the discovered path persists (a second launch doesn't need to re-scan)

---


Osiris-DevWorks (changes requested):
Thanks for fixing this. Two process fixes, plus two smaller suggestions that can land now or later.

**Fix: VERSION.TXT should say 2.2.1, not 2.3.0.** Same as the other open PRs: the installer and update checker read this file, so a wrong number breaks releases. Just remove the file from this PR.

**Fix: update your branch.** GitHub shows a conflict with the release branch.

**Suggestion:** when the scan finds nothing, nothing is saved, so the next call walks all 26 drive letters again. A disconnected network drive can make one of those checks hang for seconds. Remembering the miss until the app restarts would keep things fast for users with no install.

**Suggestion:** `get_game_install_path()` saves `root / channel` without checking that the channel folder exists. The scan only proves some channel folder exists, not the active one. A quick `is_dir()` check before saving avoids storing a path that the next read will throw away.

Stealrull:
Addressed all four items:
- VERSION.TXT now reads 2.2.1
- Branch updated to current release/2.2.1
- The scan now caches a "found nothing" result in-memory for the process's lifetime, so a no-install profile doesn't re-walk all 26 drive letters on every call (e.g. every channel switch)
- get_game_install_path() now only persists the guessed root/channel path when that channel folder actually exists (Path.is_dir()); it still returns the best-guess path either way, it just won't cache an unconfirmed guess for a later read to trust

Osiris-DevWorks (approved):
Both suggestions landed better than asked: the scan cache has a sentinel, per-test resets, and a test proving the cache answered; the is_dir() guard has a clear why-comment. Approving.

PR #277, opened 2026-07-17
[#273] Fix Generate Enhancements button stuck disabled after channel switch

### LINKED ISSUE #273: Switching between LIVE/PTU/HOTFIX even if dataforge is extracted it says it isn't and can't apply changes.
If you extract the dataforge from live but then switch to PTU or HOTFIX and extract their dataforge and apply to game then switch to another smart citizen thinks the dataforge isn't extracted and applying changes won't apply anything unless you untick and retick something then apply.

Stealrull:
## Summary
- Fixes the reported "switching between LIVE/PTU/HOTFIX thinks DataForge isn't extracted and applying does nothing until you untick/retick something". Closes #273.
- Root cause: `EnhancementsTab._enhancements_dirty` — the sole gate on the Generate Enhancements button's enabled state — was computed from disk state exactly once at tab construction, then only nudged by discrete UI events (checkbox toggles, clicking Generate itself). A channel switch changes what "needs regeneration" even means, but the flag was left holding the previous channel's value: if that channel ended clean, the button stayed disabled for a new channel that genuinely needed work, silently swallowing clicks. Retoggling any category checkbox force-sets the flag — exactly the reported workaround.
- Fix: new public `EnhancementsTab.refresh_enhancements_dirty_state()` (same shape as the existing `mark_enhancements_dirty()`) re-derives the flag from the new channel's real disk state, called from `MainWindow._on_channel_changed` right alongside the existing status-label refresh (which was already correct — only the button state was stale).

## Testing Checklist
- [x] PR is scoped to one concern (single bug fix)
- [x] `pytest tests/` passes locally (no regressions)
- [x] Affected user-facing flow exercised manually (see below)
- [ ] User-facing strings, docs/HELP.md, docs/ABOUT.md, and src/gui/coach_mark.py reviewed for drift
- [ ] New non-exempt code has matching test coverage in tests/ — GUI wiring in main_window.py/enhancements_tab.py is exempt per the project's testing rules; manual smoke testing is the documented path
- [x] Target branch is the active release/2.2.1, not main
- [x] No direct QSettings calls, no hard-coded column indices, no QProgressBar.setValue() from workers
- [x] Not an enhancement category change
- [x] VERSION.TXT matches the active release branch

## Testing performed
Exact repro from the issue exercised on a portable build: extract DataForge on channel A, switch to B and extract+apply, switch to a third channel (and back to A) — Generate Enhancements now correctly enables/disables per the new channel's actual disk state without needing to retoggle a checkbox. Note the Simple-mode one-button flow was never affected (it re-derives freshness on every click); this bug was specific to Advanced mode's Enhancements tab.

## Post-merge QA
- [ ] Repeat the multi-channel switch cycle on a clean install with 2–3 channels extracted; confirm the button state tracks each channel correctly in both directions (needs-work → clean and clean → needs-work)

---


Stealrull:
Extended per live-testing feedback: the same stale-dirty-state class hit **Save Tag Changes** (Tag Builder) and **Apply Owned Tags** (Blueprint Tracker) on a channel switch. Commit `f7196fe` lights both on switch — neither has any per-channel record of the configs/owned-set its output was last generated with, so the honest reset is clickable-after-switch (worst case one redundant regeneration/re-weave). The 2.2.1 Test Plan item was extended to cover all three buttons (#279).

Osiris-DevWorks (approved):
The disk-derived recompute for Generate Enhancements is the honest fix, and the two force-set buttons are documented tradeoffs I agree with. Approving.

PR #278, opened 2026-07-17
[#266] Fix general Blueprint Tracker bullet-matching bugs

Stealrull:
## Summary
- Extracted the general-purpose Blueprint Tracker parsing fixes from #266's original combined branch (the fuel-nozzle/mining-laser tagging feature itself now ships separately in #280). These fixes stand on their own and benefit every existing Blueprint Tracker user, independent of the new tag types.
- Fixes reputation-tier sub-header and "(Category)" annotation parsing that silently dropped whole missions' worth of bullets (`owned_items.py`).
- Recognizes the "MULTIPLE BLUEPRINT POOLS" mission header alongside "POTENTIAL BLUEPRINTS".
- Reconciles mismatched bullet names against real item entries via a generic key-slug fallback plus a small explicit-alias table (Mining Head family: Helix/Hofstede/Arbor/Klein; fuel nozzles: Norfield/Harkin/RN-7s), and recognizes the two fuel-nozzle Name-key prefixes so those aliases actually have something to resolve against.
- Deliberately does **not** include `_EXTRA_NAME_KEY_PREFIXES` entries for the ship-mounted mining-laser keys — that's a new-tag-feature-only addition living in #280.

Closes part of #266 (the general-fix portion; the tagging feature itself ships in #280).

## Testing Checklist
- [x] PR is scoped to one concern (general Blueprint Tracker bullet-matching fixes, no new feature surface)
- [x] `pytest tests/` passes locally (91/91 in the affected test files, full suite 1275 passed)
- [ ] App launches: `python src/main.py`
- [ ] Affected user-facing flow exercised manually (Blueprint Tracker tab, live mission data)
- [x] New non-exempt code has matching test coverage in `tests/`
- [x] Target branch is `release/2.2.1`, the active patch branch, not `main`
- [x] No direct `QSettings` calls, no hard-coded column indices, no `QProgressBar.setValue()` from workers

## Testing performed
Automated: full test suite green (1275 passed, 21 skipped) on the combined local `release/2.2.1` integration branch. Manual in-app verification is still needed against a real Star Citizen install/mission data — flagging for the user's own testing pass.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

Closes #267 (the XenoThreat manual-blueprints fix rides in this PR; #270 was closed in its favor).

PR #275, opened 2026-07-17
[#266] Tags for fuel nozzles, mining lasers, and scraper modules

Stealrull:
## Summary
- Adds short/medium/long Tag Builder support for three blueprint-bearing item types that previously never got tags: **Fuel Nozzle** (`FN`/`FNoz`/`Fuel Nozzle`), **Mining Laser** (`ML`/`MineL`/`Mining Laser`), and **Scraper Module** (`SCM`/`ScrMod`/`Scraper Module`). All three are gated by the existing Components → Type toggle (opt-in, off by default) and respect the user's configured abbreviation style. Closes #266.
- Fuel nozzles and scraper modules carry no `Size:`/`Grade:`/`Class:` line, so a new base.ini-only pass (`enhancements_bare_type_tags`) tags them by their description's own `Item Type:` text — robust across CIG's multiple per-manufacturer key conventions (`item_fuelnozzle_*` vs `Nozzle_FuelGiver_*`). Mining lasers live in `ships/weapons/` with no damage breakdown, so the ship-weapon tagger gained a component-style Type+Size branch for them.
- Tags now appear in **both** places a player sees the item: its own name (String Editor / loadout screens) *and* the in-game mission POTENTIAL BLUEPRINTS bullet list — the mission-bullet weaving path (`build_scitem_lookups` → `entity_name_tags`) gained the same bare-type/mining-laser support via shared helpers so the two paths can't drift.

### Blueprint-parsing bugs found and fixed along the way (all confirmed against real game data)
- **"Awarded from X level variants" reputation-tier sub-headers** were mistaken for section boundaries, truncating the blueprint scan before any bullets were read — affected dozens of missions (Adagio salvage, Bounty Hunters Guild, Security, …), whose items were entirely absent from the Blueprint Tracker.
- **"MULTIPLE BLUEPRINT POOLS"** — a second, entirely different mission header CIG uses on 35 fixture missions — wasn't recognized at all; those missions were never scanned. Also recognizes their `Pool 1`/`Pool 2` sub-labels, and the String Editor's "BP Descriptions" filter now matches them too.
- **Bullet category annotations** (`Bendix (Fuel Nozzle)`, `5CA 'Akura' (Shield)`, `(Salvage Mod)`, …) that never appear on the real item name are now stripped during normalization, so bullets join their real tagged entries instead of creating duplicate untagged "Other" rows.
[...]

Issue #220, opened 2026-07-06
Incorporate Karolinger video guide into documentation


Osiris-DevWorks:
Incorporate Karolinger video guide into documentation

youtube.com/watch?is=ly2LvECfMQXvpk_s&v=Xo1t404gsgs&feature=youtu.be

Osiris-DevWorks:
Thanks for the suggestion. Karolinger's video guide is credited in the About tab and the README as of 2.3.0.

Shipping in v2.3.0. Verified on the tester builds by Narull and Ishikudeska. If it still looks wrong after you update, reopen with a screenshot.

PR #280, opened 2026-07-17
[#266] Tags for fuel nozzles and mining lasers

### LINKED ISSUE #266: Tags for other craftable items (Fuel nozzles, mining lasers)
Want to add tags for these as I've been asked "What is a R7's?" fair question as blueprint rewards don't say that it's a Fuel nozzle, we have tags for every other kind of component mose well make it complete.

Stealrull:
## Summary
- New feature: Tag Builder support for fuel nozzles and ship-mounted mining lasers — item types that never had bracketed component tags before.
- Recognizes their non-standard Name-key conventions in the Blueprint Tracker (`item_fuelnozzle_*`, `Nozzle_FuelGiver_*`, `item_mining_mininglaser_*`) so a generated tag actually shows up joined to the right blueprint bullet, not just in the String Editor.
- Gated on the existing Components > Type toggle, matching every other component category.
- **Scope cut (maintenance mode):** the Scraper Module `[SCM]` tag originally in this PR was retired per the maintenance-mode direction — scraper module names stay untagged everywhere; the bare-type tagging path simply never emits for them. *(Corrected in review: an earlier version of this description claimed a components-config migration and a scraper negative test that were dropped along with the SCM commits.)* The `scitem_lookups` cache version is bumped to v9 (added in review) so a regen against an unchanged DataForge can't reuse a v8 pickle — that covers both pickles baked before these entries existed and any pre-scope-cut tester pickle carrying `[SCM]` tags. A stale `[SCM]` mapping row in a tester's saved components config is inert and can be removed in the Tag Builder UI. The scraper-module ITEM recognition fixes (tier headers, "(Salvage Mod)" annotations) are unchanged.
- Supersedes #275 (closed): that PR carried this feature bundled with general Blueprint Tracker bullet-matching fixes that don't need this feature to be useful — those shipped separately in a different PR (#278). This PR is feature-only, re-based on a clean history, and additionally carries the one general fix (`Fix blueprint bullet parsing: reputation-tier sub-headers + category annotations`) that the scraper-module item regression test depends on — it's a duplicate of a commit already in #278 and will de-duplicate cleanly once that branch merges up to this one.

Closes #266.

## Testing Checklist
- [x] PR is scoped to one concern (new tagging feature, no unrelated fixes beyond the one shared dependency noted above)
- [x] `pytest tests/` passes locally (full suite 1234 passed on this branch, 16 skipped)
- [x] App launches: `python src/main.py`
- [ ] Affected user-facing flow exercised manually (Tag Builder > Components > Type, Blueprint Tracker, real mission data)
- [ ] New non-exempt code has matching test coverage in `tests/`
- [x] Target branch is `release/2.3.0`, the active feature branch, not `main`
- [x] If new enhancement category: `CATEGORY_SUBTREES` and `DATAFORGE_KEEP_SUBPATHS` both updated

## Testing performed
Automated: full test suite green on this branch and on the combined local `release/2.3.0` integration branch. Manual in-app verification against a real Star Citizen install: fuel nozzle names and mining laser tags confirmed in-game by the user during 2.3.0 test builds; scraper modules verified untagged after the scope cut.

---


Ishikudeska:
@claude review

Osiris-DevWorks (changes requested):
The feature itself is correct and respects the invariants: fuel nozzles get a Type-only `[FN]` tag and mining lasers a Type+Size tag, both gated on the existing Components → Type toggle and wired into Blueprint Tracker bullet matching. Importantly the DataForge coverage invariant is **satisfied, not violated** — `enhancements_bare_type_tags` reads only `loc`/base.ini (exempt, like medical consumables) and mining-laser tagging rides the existing `weapons` subtree walk, so no new `CATEGORY_SUBTREES` / `DATAFORGE_KEEP_SUBPATHS` entry is needed. Test coverage is present. Two things before merge:

**Scope / description**
The PR checklist ticks "scoped to one concern," but the diff carries the whole 2.2.1 bug-fix bundle (#251/#261/#255/#267/#273/#281) beyond #266 — your own added test-plan section even names it. That work is now on release via #289, so please rebase onto the current tip; the diff should collapse to just the fuel-nozzle/mining-laser tagging.

**Re-verify against the now-merged #290**
#290 merged and edits `blueprint_meta.py` mining-laser handling (the `Mining_Head` → "Other" carve-out). Your PR also touches mining-laser classification/aliasing there. There's no textual conflict against the current tip, but after rebasing please confirm no double-classification/double-tagging of the mining heads and that the aliases still resolve.

**Minor**: `test_plan.py` section title says "2.2.1 bug-fix bundle" while the branch is 2.3.0; and `generate_enhancements_ini.py`'s `build_scitem_lookups` computes `is_mining_laser` then re-runs the same `_find` in `_mining_laser_component_tag` (cheap, but redundant).

---

Stealrull:
Collapsed the diff down to just the fuel-nozzle/mining-laser tagging feature per the review. The original branch also carried the whole 2.2.1 bug-fix bundle (now already on release via #289), a Scraper Module tag that was added then retired within the same branch's own history (net zero, so both commits were dropped rather than replayed and reverted), and a duplicate of the already-merged #281 fuel-nozzle-name fix (patch-id matches `70fba73`/`7d5eb82`, both already on release via #283).

Re-verified against #290's mining-head classification carve-out: this PR's Type+Size mining-laser tag operates on a different entity shape (`SEntityComponentMiningLaserParams`) than #290's `Mining_Head` loc-key carve-out, and both regression tests still pass, so there's no double-classification.

Full test suite passes.

Stealrull (commented):
## Self-review (`/review_all`)

Full file-by-file pass on all 8 files (`generate_enhancements_ini.py` — the largest, ~500 lines — `blueprint_meta.py`, `tag_builder.py`, and 5 test files).

**Finding, fixed in `a7a260b`:** The `"Fuel Nozzle"` mapping entry's comment in `tag_builder.py` claimed the generator "always forces this element on for them" — the opposite of the actual behavior. Confirmed against four independent sources before touching anything: the real code (`_bare_type_tag_from_desc` returns `None` when the user's Type element isn't enabled), the comment on that same function elsewhere in `generate_enhancements_ini.py` ("they don't get force-shown"), and two separate test docstrings (`test_bare_type_tags.py`'s `test_no_tag_when_type_disabled_by_default`, and `test_ship_weapon_tag.py`'s `TestShipWeaponTagMiningLaser` class docstring for the sibling feature) both stating "opt-in, not force-shown." Rewrote the comment to match what the code and tests actually do.

**Verified clean** (this is a large, carefully-built feature):
- Mining laser tagging is correctly narrowed to only real mining lasers via the `SEntityComponentMiningLaserParams` check — with a dedicated regression test confirming a combat weapon in the same `weapons/` directory still gets no tag, guarding against reopening #220.
- The `name_fallback_tags` refactor (moving tag application outside the tier-1/2/3 name-resolution chain, so a tag can apply "regardless of which tier supplied the name") is deliberate and has its own precedence test confirming the entity-XML tag wins over the name-fallback when both exist — no double-tagging risk.
- The bare-type "Item Type:" regex correctly handles the game's `\n`-as-two-characters loc-string convention (not a real newline).
- The small explicit allow-list (`_BARE_TYPE_NAMES = frozenset({"Fuel Nozzle"})`, rather than matching any component type by its Item Type text) deliberately avoids double-tagging items already covered by the stricter Class-based path — confirmed by its own dedicated test.
- The cache-version bump (`blueprint_pools`: v12 → v13) is documented correctly so a regen doesn't silently reuse a stale pre-fix pickle.

Suite: 1359 passed, 21 skipped, same 5 pre-existing Chinese/Italian/Japanese activation failures tracked by #318, unrelated to this PR.

Osiris-DevWorks (approved):
Both review blockers are resolved: the scitem_lookups v9 bump landed with a version-history entry (stale v8 pickles, including SCM-era tester pickles, now flush on regen), and the description now matches the shipped diff. The branch update also resolved the comment-block conflict with #335, keeping both the filename-prefix block and the broader bare-type comment. Full suite passes (1524). Merging.

Issue #247, opened 2026-07-15
i18n: sweep remaining hardcoded UI strings that bypass tr()


Osiris-DevWorks:
A sweep during the 2.2.0 pre-release language check found user-facing strings in `src/gui/` that never route through `tr()`, so no translation can reach them no matter how complete the language files are. French, PT-BR, and Spanish users see these in English today.

Known offenders (sample, not exhaustive):

- The Import INI flow: dialog title, the "No Base Data" / "No Valid Keys" / "Nothing to Import" / "Import Complete" message boxes
- The error dialog window title ("Smart Citizen" + "Error")
- Status-bar messages: "Downloading INI file...", "Ready", "Starting up", "Generating enhancements in background", "Enhancements generated"
- Tag Builder element controls: the Move up / Move down / Untick-to-exclude tooltips
- "Success" backup-restored message box

The fix is mechanical but touches a lot of call sites: wrap each literal in `tr()`, add the keys to `languages/english/ui.json`, and let the pre-release backfill cover the other languages. Same shape as the #237 tooltip sweep, which got the tooltips but not these.

Out of scope for 2.2.0 (string sweep, not a bug fix). Good first target for the 2.3.0 branch.

Osiris-DevWorks:
The remaining hardcoded UI strings now route through `tr()`, so they translate along with the rest of the interface.

Shipping in v2.3.0. Verified on the tester builds by Narull and Ishikudeska. If it still looks wrong after you update, reopen with a screenshot.

PR #282, opened 2026-07-18
Fix crash: tr() kwarg named 'key' collides with its own first parameter

Stealrull:
## Summary
- Crash report: opening the "Custom..." resolution in Import INI and entering a value raises `TypeError: tr() got multiple values for argument 'key'`. `import_dialog.py`'s `custom_value_prompt` call already exists on `release/2.2.1` as `tr("import_dialog.custom_value_prompt", key=key)` — passing a loc key as an interpolation variable is natural, but `tr(key: str, **kwargs)` has its own first parameter also named `key`, so Python rejects the call before the function body ever runs.
- Renames `tr()`'s own parameter to `i18n_key` so `key` is free for any caller, and fixes the one existing crashing call site plus its `{key}` → `{loc_key}` placeholder across all four locale files.
- Scoped narrowly to this one pre-existing crash. A second, unrelated `tr()`-with-`key=` collision in `main_window.py`'s "Copy Key" action only exists once the #247 hardcoded-string sweep converts that line from a plain f-string to a `tr()` call — that lands alongside the sweep itself in a separate PR, not here.

## Testing Checklist
- [x] PR is scoped to one concern (single crash fix)
- [x] `pytest tests/` passes locally (full suite 1205 passed, 16 skipped)
- [ ] Affected user-facing flow exercised manually (Import INI > Custom... resolution)
- [x] New non-exempt code has matching test coverage in `tests/`
- [x] Target branch is `release/2.2.1`, the active patch branch, not `main`

## Testing performed
Automated: full test suite green. Manual verification of the Import INI Custom... flow is still needed — flagging for the user's own testing pass.

---

### DIFF
--- a/languages/english/ui.json
+++ b/languages/english/ui.json
@@ -1055,7 +1055,7 @@
       "at": ""
     },
     "custom_value_prompt": {
-      "ht": "Enter custom value for key:\n{key}",
+      "ht": "Enter custom value for key:\n{loc_key}",
       "at": ""
     },
     "auto_add_summary": {
--- a/languages/french/ui.json
+++ b/languages/french/ui.json
@@ -1029,8 +1029,8 @@
       "at": "Valeur personnalisée"
     },
     "custom_value_prompt": {
-      "ht": "Entrez une valeur personnalisée pour la clé :\n{key}",
-      "at": "Entrez une valeur personnalisée pour la clé :\n{key}"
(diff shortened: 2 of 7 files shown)


Osiris-DevWorks (approved):
Verified the rename is safe: one key= call site and one {key} placeholder exist in the codebase, both fixed here, and the locale edits are mechanical placeholder renames. Approving.

PR #289, opened 2026-07-22
Merge release/2.2.1 into release/2.3.0

Stealrull:
## Summary
`release/2.3.0` was branched before the full 2.2.1 bug-fix bundle (#251, #255, #259, #261, #267, #269, #273, #281) merged, so every 2.3.0 feature/fix branch built on top of it was silently missing all of it. Most visibly: a fresh 2.3.0 portable build today didn't auto-detect the Star Citizen install path at all — the #269 fix simply wasn't there yet.

This brings `release/2.3.0` fully current with `release/2.2.1` so nothing built on top of it going forward repeats this. Clean merge, no conflicts. `VERSION.TXT` corrected back to `2.3.0` after the merge (git took `2.2.1`'s value since `release/2.3.0` never independently touched that file after branching).

## Testing Checklist
- [x] PR is scoped to one concern (housekeeping merge, no new behavior)
- [x] `pytest tests/` passes locally (1276 passed, 16 skipped)
- [x] Target branch is `release/2.3.0`
- [x] `VERSION.TXT` matches the active release branch (2.3.0)

## Testing performed
Full suite green after the merge and after the VERSION.TXT correction.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #294, opened 2026-07-23
[#293] Fix portable zip paths too long for Recycle Bin

### LINKED ISSUE #293: Portable version, folder lengths too long for recycle bin
Currently if you delete the portable version you get the windows error that the file names are too long for the recycle bin and will be permanently deleted, is a easy fix to this like the previous versions fix so will add it (I noticed it a lot as I run test versions of every bug so it got on my nerves)

Stealrull:
Fixes #293

## Problem
Deleting an extracted portable build gives the Windows "file/folder names are too long for the Recycle Bin, will be permanently deleted" error.

## Root cause
Two things were stacking extra path length on top of the portable build's already-deep `patches/` tree:
- The zip wrapped its contents in a top-level `SmartCitizen-Portable-vX.Y.Z/` folder. Windows Explorer's "Extract All" already creates an enclosing folder of that name, so the zip's own wrapper doubled the prefix on the common extraction path.
- `--collect-all=lxml` pulls in lxml's bundled `isoschematron` reference-XSLT resources, which nothing in this codebase uses (no `isoschematron`/`schematron` references anywhere) — just unused files adding nested path depth.

Combined, these were enough to push some extracted file paths past Windows' 260-char `MAX_PATH`, which is exactly when Explorer can't compute a Recycle Bin move and falls back to permanent delete.

## Fix
In `scripts/build/build_exe.py`:
- Stop wrapping the portable zip's contents in the top-level version folder.
- Strip the unused `_internal/lxml/isoschematron/resources/` folder from the frozen build after PyInstaller runs.

## Testing
Built a portable with these changes and confirmed:
- Zip's top-level entries are the `.exe` and `_internal/` directly (no double-nesting when extracted with "Extract All").
- The isoschematron resource files are gone from the build (391 vs 400 files previously), with no change to the app's actual XML parsing.
- Worst-case extracted path length dropped from what would've been ~197 chars (Desktop) / ~232 chars (network share) down to 167 / 202 — well clear of the 260-char limit.

Reporter confirmed on a real portable build that the Recycle Bin warning is gone and the app still runs normally.


Osiris-DevWorks (dismissed):
Approving. The `build_exe.py` fix is correct and well-commented.

- Dropping the redundant top-level `SmartCitizen-Portable-vX.Y.Z/` zip wrapper (Explorer's "Extract All" already creates it) plus stripping the unused `isoschematron` resources shortens the worst-case extracted path below the 260-char `MAX_PATH` that triggers the Recycle Bin permanent-delete fallback. Path-length math checks out; `isoschematron` is confirmed unused in the tree.

The 2.2.1-bundle scope is resolved now that #289 has merged. Minor follow-up (non-blocking): the `.github/workflows/portable-preview.yml` comments still describe the old top-level-folder behavior — worth a refresh since this PR changes what they describe.

---

PR #295, opened 2026-07-23
[#292] Clear Generate Enhancements dirty flag on every pipeline trigger

### LINKED ISSUE #292: "Generate Enhancements" Remains Red after Auto-Generation
The Generate Enhancements button shows red text when there are unsaved changes since last generated enhancements. However, it does not detect when smart citizen's initial automated workflow (for example after a new version) has generated enhancements. It only detects manually triggered ones.

We should update this to detect the automated generation.

Stealrull:
Fixes #292

## Problem
The Generate Enhancements button's red "needs regeneration" indicator only got cleared by the manual button click handler. It never cleared after Smart Citizen's own automated generation (e.g. the startup/post-extraction freshness check), so the button stayed red even though enhancements had just been successfully regenerated.

## Root cause
`_on_generate_enhancements_clicked()` cleared the dirty flag itself, ad hoc, right when the button was clicked. But every trigger for enhancement generation — the manual click, Tag Builder apply, the simple-mode run button, and the automated freshness check — funnels into `MainWindow._run_enhancements_pipeline()`, and only the click path went through the code that cleared the flag. The shared completion handler (`set_operation_idle()`) only re-applies whatever the dirty flag already is on a successful run (by design — a setting changed mid-run should still show dirty afterward) and never forces it clean on its own.

## Fix
- Added `EnhancementsTab.mark_enhancements_clean()`, a public counterpart to the existing `mark_enhancements_dirty()`.
- Call it once, at the top of `_run_enhancements_pipeline()` — the true common entry point for every trigger — instead of ad hoc in the click handler.

Deliberately **not** added to `_run_enhancements_generation()` itself: that function also handles per-language regeneration for the UI language switcher, which writes to a different per-language cache directory than the one the button's dirty state tracks. Clearing there unconditionally would wrongly mark the button clean after a non-English prose regen while the actual English/base cache stayed stale.

## Testing
Built a portable with this change. Reporter confirmed on a real run that after a channel with missing/stale enhancement files triggers the automated freshness-check generation on launch, the Generate Enhancements button comes up clean (not red) afterward — same as the existing manual-click behavior, and toggling a setting afterward still correctly lights it back up.


Osiris-DevWorks (dismissed):
Approving. Correct and complete for #292.

- Moves the dirty-flag clear out of the click handler into `_run_enhancements_pipeline()`, the single common entry point, via `mark_enhancements_clean()`. I confirmed all trigger paths funnel through it (manual click signal, the automated freshness check, and the simple-mode chain), so the automated regeneration now clears the button too.
- No double-clear (same-thread signal), failure re-sets dirty via `set_operation_idle(success=False)`, and the deliberate exclusion of the per-language `_run_enhancements_generation()` is sound. Consistent with the sibling `mark_owned_clean()` pattern (#297).

The 2.2.1-bundle scope is resolved now that #289 has merged.

---

PR #297, opened 2026-07-23
[#296] Clear Apply Owned Tag dirty flag after a log scan applies it

### LINKED ISSUE #296: "Apply Owned Tag" remains red after scan
Same kind of issue that was reported for automated enhancements currently the blueprint scanner says it's applied the owned tag after a scan but the apply owned tag button remains red.

Image

Stealrull:
Fixes #296

## Problem
After a blueprint log scan finds new owned items, the summary dialog tells the user their tags are now visible ("N are visible in your current blueprint lists now") — but the "Apply Owned Tags" button immediately shows red, as if nothing had been applied.

## Root cause
Same class of bug as #292 (Generate Enhancements). `MainWindow._finish_blueprint_scan_queue()` already calls `_recompute_owned()` to weave the newly-scanned items' `[Owned]` tags into the currently loaded blueprint bullets — the exact same re-weave the "Apply Owned Tags" button itself triggers — but immediately afterward called `self.blueprint_tracker_tab.mark_owned_dirty()`, lighting the button back up right after the real work had already happened.

## Fix
- Added `BlueprintTrackerTab.mark_owned_clean()`, a public counterpart to the existing `mark_owned_dirty()`.
- `_finish_blueprint_scan_queue()` now calls `mark_owned_clean()` instead of `mark_owned_dirty()` right after `_recompute_owned()`.

Deliberately **not** touching the channel-switch call to `mark_owned_dirty()` (`main_window.py` ~3355) — per its own comment, that one is intentionally dirty by design: the owned set's items may not all be visible bullets in the newly-switched channel's data yet, so the button is meant to stay lit as an easy re-weave affordance. The scan path is different — the weave already fully happened for what's currently loaded, so there's nothing left to prompt for.

## Testing
Built a portable with this change (on top of the #293 and #292 fixes already in `release/2.3.0`'s pipeline). Reporter confirmed on a real scan that finds new blueprints that the Apply Owned Tags button now stays clean/grey right after the scan, consistent with the summary dialog.


Osiris-DevWorks (dismissed):
Approving. Correct, minimal, and consistent with the sibling dirty-flag fixes.

- After a blueprint log scan, `mark_owned_clean()` routes through the single `_set_owned_btn_dirty()` chokepoint (same path as the Apply button), so the button greys out correctly instead of re-dirtying via the erroneous `mark_owned_dirty()` call. No double-clear; the flag is only re-set by the tab's own arrow-move methods.
- GUI wiring, test-exempt per the coverage policy; behavior confirmed on a real scan.

The 2.2.1-bundle scope that inflated the original diff is resolved now that #289 has merged.

---

PR #302, opened 2026-07-23
Give Save Tag Changes a real channel-switch freshness check

Stealrull:
## Problem
On channel switch, Generate Enhancements already does a real freshness check (`refresh_enhancements_dirty_state`) and only lights up when the new channel genuinely needs regeneration. Save Tag Changes instead lit up unconditionally on every switch, since nothing recorded which Tag Builder config a channel's generated INIs were built from.

## Fix
Added a per-channel/per-language `.tag_config_stamp` file, mirroring the existing `.dataforge_stamp`:
- `tag_builder.tag_config_fingerprint()` — deterministic sha256[:12] of the tag configs (`TagConfig.to_json()` is already canonical) plus the annotate-mission-descriptions toggle.
- The generator worker writes the stamp alongside `set_enhancements_stamp`, from the configs it actually generated with.
- `refresh_tag_builder_dirty_state()` now lights the button only when the channel's stamp is missing or differs from a fingerprint of the **live** Tag Builder pages (not just persisted settings) — so unsaved in-session edits still correctly keep the button lit across a switch. A missing stamp reads dirty, so pre-existing channels behave exactly as before until regenerated.

I also looked at giving Apply Owned Tag (Blueprint Tracker) the same treatment, since its `[Owned]` weave always re-runs on reload and technically has no per-channel staleness to detect. I implemented that too, but reverted it — the button staying clickable after a switch is a deliberate, wanted affordance (force-reapply owned tags to the newly-selected channel), not a bug, so that behavior is unchanged from current `release/2.3.0`.

## Testing
Adversarially reviewed (4 independent review passes + skeptic verification on each finding) before landing — confirmed no false-clean/false-dirty holes, including the async-extraction deferred-reload path. Added 13 new/updated tests (fingerprint determinism/order-independence/sensitivity, stamp round-trip/isolation/corrupt-stamp handling). Built a portable and confirmed in the real app: Save Tag Changes stays grey across a channel switch when nothing changed, lights up when the config genuinely differs, and unsaved edits survive a switch.


Osiris-DevWorks (dismissed):
Approving. The freshness-check feature is correct, well-tested, and standards-clean.

- Per-channel/per-language `.tag_config_stamp` mirrors `.dataforge_stamp`; the stamp lives under `get_enhancements_dir()`, so a switch reads the *target* channel's stamp — it genuinely detects the switch.
- Live-page fingerprint keeps unsaved edits lit across a switch; a missing stamp reads dirty, so un-stamped channels behave exactly as before.
- Worker stamp, live fingerprint, and `get_current_tag_config_fingerprint()` iterate the same `CATEGORIES` set — no spurious mismatch. Fingerprint inputs are pre-fetched on the main thread and handed to the worker (honors the tag-config hand-off rule); corrupt-stamp guard matches the existing pattern.
- Deliberate Apply-Owned-Tag exclusion is documented and correct.

The bundled 2.2.1 noise that inflated this diff is resolved now that #289 has merged.

---

PR #303, opened 2026-07-23
Show a clear message when Data.p4k is locked by the RSI Launcher

Stealrull:
Stacks on #252 (i18n hardcoded-string sweep) — this uses the `extract.*` i18n keys and tr()-based error dialogs that PR introduces; the extraction error dialogs are still hardcoded English on plain `release/2.3.0`.

## Problem
Extracting the base.ini or DataForge cache while the RSI Launcher is downloading/verifying a game update failed with a raw .NET stack trace: `unp4k.exe exited with code 3762504530 ... System.IO.IOException: The process cannot access the file '...Data.p4k' because it is being used by another process.` The launcher holds an exclusive lock on `Data.p4k` during a patch, and unp4k opens it directly.

## Fix
- The exit code (`0xE0434352`) is just the generic .NET managed-exception code, so detection keys off the message text `"being used by another process"`.
- Added `P4kLockedError(RuntimeError)` so both unp4k call sites (`extract_global_ini`, `extract_dataforge`) can raise a distinct, typed exception for this case instead of a plain `RuntimeError`.
- When matched, raises `P4kLockedError` with a plain, actionable message: "Star Citizen appears to be updating right now. Its game files (Data.p4k) are locked by another program, usually the RSI Launcher downloading or verifying an update. Wait until all updates and verification finish in the RSI Launcher, then try again." New i18n key `extract.p4k_locked` (English text; empty FR/PT-BR/ES stubs fall back to English).
- **Avoiding a duplicate popup:** `P4kExtractWorker`/`DataForgeExtractWorker` each wrap their extraction call in a blanket `except Exception: logger.exception(...)`, which logs at ERROR unconditionally and independently triggers MainWindow's global `ErrorDialogHandler` (any ERROR-level log record, app-wide) — so the friendly dialog was popping up *twice*, once from the worker's generic handler and once from the intended one. Both workers now catch `P4kLockedError` before the generic handler and log it at `WARNING` instead (full detail is already logged by `_raise_unp4k_failure`) — the `error` signal still fires and the friendly dialog still shows, just without the duplicate. Any other unp4k failure is still a plain `RuntimeError` and still correctly triggers the generic handler.

## Testing
Built portables through several iterations of this fix (the duplicate-dialog bug took two passes to actually fix — the log-level downgrade inside the extraction helper alone wasn't enough, since the workers' own blanket exception handler independently logged at ERROR). Reporter confirmed on a real RSI Launcher update that the locked-file scenario now shows exactly one friendly dialog. 36 tests in `test_pak_extraction.py` (added: typed-exception assertions, log-level assertions for both the locked and non-locked cases).

### DIFF
--- a/languages/english/ui.json
+++ b/languages/english/ui.json
@@ -2337,7 +2337,7 @@
       "at": ""
     },
     "p4k_locked": {
-      "ht": "Star Citizen appears to be updating right now — its game files (Data.p4k) are locked by another program, usually the RSI Launcher downloading or verifying an update.\n\nWait until all updates and verification finish in the RSI Launcher, then try again.",
+      "ht": "Star Citizen appears to be updating right now. Its game files (Data.p4k) are locked by another program, usually the RSI Launcher downloading or verifying an update.\n\nWait until all updates and verification finish in the RSI Launcher, then try again.",
       "at": ""
     }
   },
--- a/languages/english/ui.json
+++ b/languages/english/ui.json
@@ -2335,6 +2335,10 @@
     "extraction_error_title": {
       "ht": "Extraction Error",
       "at": ""
+    },
+    "p4k_locked": {
+      "ht": "Star Citizen appears to be updating right now — its game files (Data.p4k) are locked by another program, usually the RSI Launcher downloading or verifying an update.\n\nWait until all updates and verification finish in the RSI Launcher, then try again.",
+      "at": ""
     }
   },
   "apply": {
--- a/languages/french/ui.json
+++ b/languages/french/ui.json
@@ -2335,6 +2335,10 @@
     "extraction_error_title": {
       "ht": "",
       "at": "Erreur d'extraction"
+    },
+    "p4k_locked": {
+      "ht": "",
+      "at": ""
     }
   },
   "apply": {
--- a/languages/portuguese_br/ui.json
+++ b/languages/portuguese_br/ui.json
@@ -2335,6 +2335,10 @@
     "extraction_error_title": {
       "ht": "",
       "at": "Erro de extração"
+    },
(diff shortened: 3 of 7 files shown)


Stealrull (commented):
**What:** Adds a typed `P4kLockedError` and detects when the RSI Launcher has `Data.p4k` locked during an update, showing one friendly translated message instead of a raw .NET stack trace shown twice (the duplicate-dialog bug). Detection keys off the exception's message text (`"being used by another process"`), not the exit code, which is just the generic .NET managed-exception code.

Reviewed all 7 files: exception typing and thread-safe signal-only emits in `workers.py` are correct (subclass caught before the generic handler, logged at WARNING to avoid double-triggering the global `ErrorDialogHandler`), the case-insensitive detection in `pak_extractor.py` is clean, and `tests/test_pak_extraction.py` has genuinely good coverage — including a `caplog`-based regression guard locking the WARNING-not-ERROR log level that's the actual fix. No correctness issues found.

**Base branch:** This PR targets `issue/247-i18n-hardcoded-string-sweep`, which is dead — its own PR (#252) closed unmerged, superseded by #324 which landed the i18n sweep directly on `release/2.3.0`. The `extract.*` i18n keys this PR depends on already exist on tip. I test-cherry-picked all 4 commits onto current `release/2.3.0`: zero conflicts, full suite passes. Retargeting now.

No blockers otherwise.

Osiris-DevWorks (approved):
Verified the lock-detection logic, i18n wiring (English key with fallback for the others, per policy), worker threading, and the 8 new tests on _raise_unp4k_failure. One cosmetic nit, not blocking: the PR body says 36 tests in test_pak_extraction.py, the file has 34. Merging.

PR #254, opened 2026-07-16
[#248] i18n: backfill FR/PT-BR docs for 2.2.0; add Spanish HELP/ABOUT/LEGAL

Stealrull:
## Summary
- **French & PT-BR** `HELP.md`/`ABOUT.md` predated the 2.2.0 cycle. Backfilled every missing section: Simple & Advanced mode, dirty-state button colors + the Unapplied Changes close prompt, App Updates, the Blueprint Tracker tab, Mission Titles, the FAQ tab, medical consumables / RS tags / rep-track XP, Restore user.ini, and the credits updates (Stealrull, hkstrongside, Thord82, XaileiShiv, Mindbulletz, Spanish in the language list). Their `LEGAL.md` was already current, so it's untouched.
- **Spanish** had no translated docs at all — the app silently fell back to English even though its UI is fully covered. Added all three (`HELP.md`/`ABOUT.md`/`LEGAL.md`), translated in full from the English originals. They're picked up automatically by `AppSettings.get_localized_doc_path`'s `languages/<lang>/<file>` resolution — no code change needed.
- **Terminology matches the shipped UI**, not the English rename: every button/tab name in the docs was cross-checked against each language's `ui.json` (`Aplicar mejoras`, `Rastreador de blueprints`, `Suivi des plans`, `Aplicar Aprimoramentos`, …). That includes deliberately using French's stale `Appliquer les modifications d'étiquettes` — it's what the French app shows today per the `apply_tag_changes_btn` re-review note in `TRANSLATIONS.md`, and docs should describe the screen the user actually sees.
- **Provenance**: all AI-drafted per the issue's plan, logged in `TRANSLATIONS.md`'s backfill log, and flagged for review by the language leads — **Akwa/Ishikudeska (français), Nxzzin (português brasileiro), Thord82 (español)** — rather than silently shipping. These docs were already AI-provenance for FR/PT-BR (the human `ht` tracking applies to `ui.json` only), so no human translation was overwritten.
- **Structure verified against English**: `##` section counts match across all three languages for all three docs (29/12/6), total bullet counts match in ABOUT (73), and no stale pre-2.2.0 button names remain.

Closes #248.

## Testing Checklist
- [x] PR is scoped to one concern (translated-docs backfill)
- [x] `pytest tests/` passes locally (docs-only change; no code touched)
- [ ] App launches: `python src/main.py`
- [ ] Affected user-facing flow exercised manually (switch language → open Help/About/Legal panels)
[...]

PR #304, opened 2026-07-23
Move the frontend version watermark to its own line

Stealrull:
Part of #319 (the main-menu watermark line-break half; the mission-details `+` removal is tracked separately).

## Problem
The main-menu watermark ("Localizations Enhanced with Smart Citizen v#") was appended after " | " on the same line as the build info ("Alpha 4.9: Frontier Tensions"), crowding a single line.

## Fix
Switched the separator to a literal `\n` — the same loc-string line-break convention already used for the journal entry stamp (`_JOURNAL_STAMP_RE`) — so it renders on its own second line instead:

```
Alpha 4.9: Frontier Tensions
Localizations Enhanced with Smart Citizen v2.3.0
```

`_FRONTEND_VERSION_STAMP_RE` now matches both the new `\n`-separated form and the original `" | "`-separated form (pre-2.3.0 installs already have that on disk), so re-applying replaces either cleanly instead of double-stamping or leaving the old one dangling.

Updated `docs/HELP.md`, `src/gui/CLAUDE.md`, and the French/Portuguese-BR translated `HELP.md` copies (identical inline code literal, not translated prose). Spanish `HELP.md` doesn't exist yet on this PR's base — see the small follow-up PR targeting #254 for that one.

## Testing
13 tests in `test_frontend_version_stamp.py` (existing coverage updated for the new default output, plus a new case covering old-pipe-format-on-disk → new-format rollover, and new-form entries in the whitespace-tolerance regex test). Built a portable and confirmed in-game: the watermark renders on its own second line in the Star Citizen main menu.


Osiris-DevWorks (dismissed):
Approving. The watermark change is correct and the stamp round-trip is verified.

- Separator changed from `" | "` to a literal `\n` so the "Localizations Enhanced with Smart Citizen v{VERSION}" line sits on its own line. The updated regex strips both the new `\n`-separated form and the legacy `" | "` form before re-appending, so there's no stamp bypass, no double-stamp, and pre-2.3.0 pipe watermarks roll forward cleanly. `.rstrip()` won't eat the literal `\n`.
- The stamp-regex test lock is updated (default-output flipped to `\n`, a roll-forward-from-pipe case added, whitespace-tolerance cases added); docs synced.

Implements the main-menu watermark half of #319 (the rep-reward `+` half landed as #320). The 2.2.1-bundle scope is resolved now that #289 has merged.

---

PR #272, opened 2026-07-17
[#268] feat: option to also scan LIVE/HOTFIX for owned blueprints

Stealrull:
## Summary
- Players with more than one install (LIVE + HOTFIX, which share the same account progression) could earn a blueprint on one channel and never see it marked owned, since "Scan Logs for Owned Blueprints" only ever covered the active channel.
- Adds a checkbox to the Blueprint Tracker tab: when on, the scan also covers whichever of LIVE/HOTFIX isn't currently active, if it's installed. PTU/EPTU/TECH-PREVIEW are never included — separate test builds, not the same progression.
- `src/utils/settings.py`: new `scan_other_channels` setting; generalized the per-channel BP Scan watermark functions to accept an explicit `channel` parameter (already stored per-channel, just had no way to target a channel other than the active one).
- `src/gui/main_window.py`: new pure `_channels_to_scan()` helper plus a sequential scan queue that runs each channel through the existing single-channel worker unchanged, then shows one combined summary once the queue drains.

Closes #268.

## Testing Checklist
- [x] PR is scoped to one concern (single feature)
- [x] `pytest tests/` passes locally (1199 passed, 16 skipped on this branch)
- [x] App launches: confirmed via portable build
- [x] Affected user-facing flow exercised manually: confirmed by the reporting user on a real portable build with both LIVE and HOTFIX installed — scans both in one pass with a single combined summary
- [x] User-facing strings, `docs/HELP.md`, `docs/ABOUT.md`, and `src/gui/coach_mark.py` reviewed for drift (`HELP.md`'s Blueprint Tracker section updated with the new checkbox's behavior)
- [x] New non-exempt code has matching test coverage in `tests/` (`_channels_to_scan()` — the one pure/testable piece of this feature — covered by 8 new tests in `test_blueprint_scan_channels.py`; the generalized watermark/checkbox settings covered by 4 new tests in `test_owned_items.py`)
- [x] Target branch is the active `release/2.3.0`, not `main`
- [x] No direct `QSettings` calls, no hard-coded column indices, no `QProgressBar.setValue()` from workers
- [x] `VERSION.TXT` matches the active release branch

## Testing performed
Full suite (1199 passed, 16 skipped). Built a portable v2.3.0 test build; the reporting user confirmed on their real install (both LIVE and HOTFIX present) that toggling the checkbox on scans both channels in one pass and shows a single combined summary, and that leaving it off (or only one of the pair installed) behaves exactly as before.

## Post-merge QA
[...]

Issue #268, opened 2026-07-17
Have an option (tickbox) for smart citizen blueprint scanner to check for other installs logs (hotfix)


Stealrull:
A lot of players have more then one install due to hotfix channel, having this option would make the blueprint scan complete for if something was earnt on the hotfix channel but not on the live install.

Osiris-DevWorks:
Added. The scan now also reads whichever of LIVE and HOTFIX is not your active channel, since those two share account progression. It is on by default. PTU, EPTU, and TECH-PREVIEW have separate progression and are never scanned.

Shipping in v2.3.0. Verified on the tester builds by Narull and Ishikudeska. If it still looks wrong after you update, reopen with a screenshot.

PR #310, opened 2026-07-24
[#308] Add a "Rescan all logs" checkbox to force a full blueprint log rescan

### LINKED ISSUE #308: Add a re-scan blueprint logs tickbox
Ran into a couple of users who have asked this they have accidently changed something with blueprints and want to rescan the logs, a simple tickbox would be easy to add.

Stealrull:
Closes #308.

## Problem

"Scan Logs for Owned Blueprints" only reads events newer than the last scan's watermark (#222), which keeps repeat scans fast but leaves no way to recover if a user's Owned set ends up wrong -- e.g. an accidental unown, or a scan that ran before a log file finished being written.

## Fix

Adds a **"Rescan all logs (ignore last scan)"** checkbox on the Blueprint Tracker tab, in the same row as the existing "Also scan LIVE/HOTFIX" checkbox (#268). When checked, the next scan bypasses the saved watermark for every queued channel and re-walks each log back to the scanner's own March-2026 epoch floor.

One-shot by design: MainWindow unchecks it once the whole queued scan finishes (success or error), so it doesn't silently keep forcing a slow full rescan on every future click. Deliberately not persisted to `AppSettings`, unlike "also scan other channels" -- this isn't a standing preference.

New pure helper `_blueprint_scan_since(force_rescan, watermark)` decides the effective watermark per queued channel, mirroring the existing `_channels_to_scan()` pattern for testability.

## Depends on #272

This branch is built on top of `issue/268-scan-other-channels-for-owned-blueprints` (#272, still open) since the new checkbox shares a layout row with that PR's checkbox. That branch isn't on the upstream repo, so this PR is based on `release/2.3.0` like #272 itself -- **the diff below includes #272's changes too** until #272 merges. Please review/merge #272 first (or review this one together with it); once #272 lands, this diff will shrink to just the #308 checkbox.

## Testing

`tests/test_blueprint_force_rescan.py` (4 cases: forced vs unforced, with and without an existing watermark). Full suite passes. Built a portable combined with every other in-flight 2.3.0 branch and confirmed in-app: checking the box and running a scan re-reads every log from scratch instead of just new events, and the checkbox unchecks itself once the scan completes.


Osiris-DevWorks (commented):
The core #308 logic is correct and thread-clean: `_blueprint_scan_since` returns `None` on force, the scanner treats `since=None` as the epoch floor (so it genuinely forces a full rescan), the checkbox is read once on the main thread before the queue starts, and `reset_force_rescan_checkbox()` runs on both success and error paths. Nice. A few things to tidy before it lands, none blocking on their own:

- **`retranslate_ui()` omits the new checkbox** (`src/gui/blueprint_tracker_tab.py`) — it re-translates `_scan_other_channels_checkbox` but not `_force_rescan_checkbox`, so on a runtime language switch the force-rescan text/tooltip won't update. Add the matching `setText`/`setToolTip`.
- **Empty translations** — `force_rescan_checkbox` / `_tooltip` were added to `languages/{french,spanish,portuguese_br}/ui.json` with both `ht` and `at` empty. English fallback keeps it working, but each activated language needs at least an AI translation (this is also the current merge conflict against the tip — those three files — so fold it into the rebase).
- **Docstring pointer** — `_blueprint_scan_since` says "see test_blueprint_scan_channels.py" but its tests live in `test_blueprint_force_rescan.py`.
- **HELP.md** — the new "Rescan all logs" checkbox isn't documented (the added bullet covers #272's other-channels toggle instead). Low urgency since the tooltip covers it in-app.

Scope is resolved now that #289 has merged — please rebase (the only conflicts are the three `ui.json` files above). Coordinate with #272, which touches the same tab layout row.

---

Stealrull (commented):
## Self-review (`/review_all`)

Full file-by-file pass on all 11 files (settings.py, blueprint_tracker_tab.py, main_window.py, three test files, docs, and 4 i18n files).

**Inherited, already fixed elsewhere — not a #310-specific issue:** This branch is built on top of #272, so its `settings.py` diff still carries the "Off by default" stale comment I already caught and fixed there (`56bee53` on #272). It'll resolve on its own once this PR rebases onto #272's current tip; nothing to fix on this branch specifically.

**Finding, fixed in `9fa35dd`:** `_bp_scan_force_rescan` was the only one of the four multi-channel-scan state attributes not initialized in `setup_ui()` — its three siblings (`_bp_scan_queue`, `_bp_scan_channel`, `_bp_scan_new_names`) all are. Not a live bug (`_run_blueprint_log_scan()` always sets it before any code path reads it, so there's no real `AttributeError` risk), but worth closing the inconsistency for its own sake.

**Verified clean:**
- `_blueprint_scan_since()` is a clean pure function with full test coverage (forced/unforced × watermark-present/None).
- The force-rescan flag is read exactly once, at the start of the whole multi-channel run (not per-worker), so a checkbox toggle mid-scan can't retroactively affect a channel already in flight.
- The one-shot checkbox reset (`reset_force_rescan_checkbox()`) fires unconditionally at the top of `_finish_blueprint_scan_queue()`, before the "nothing new found" early return — so it resets whether or not the run turned up anything, matching the stated one-shot intent.
- `retranslate_ui()` correctly updates both new checkboxes' text/tooltip on a language switch.
- French/Spanish/Portuguese translations are accurate and consistent with the English source.

Suite: 1346 passed, 21 skipped, same 5 pre-existing Chinese/Italian/Japanese activation failures tracked by #318, unrelated to this PR.

Osiris-DevWorks (dismissed):
The #308 change itself verified clean in review: force flag read on the main thread, pure _blueprint_scan_since with tests, checkbox reset covers both finish and error paths, retranslate covered, and the watermark's max() write means a forced rescan can't corrupt it. The stacked #272 content is now merged, so this diff is down to the #308 change. Merging.

Osiris-DevWorks (approved):
Re-approving: the previous approval was dismissed for a merge-base change (#272 landing), which does not affect this PR's content. Verification unchanged: force flag read on the main thread, pure _blueprint_scan_since with tests, checkbox reset covers finish and error paths, watermark max() write is corruption-safe.

Issue #215, opened 2026-07-02
Disabled "CF" Label Still Appears


Osiris-DevWorks:
Since I decided to use the component labels on commodities, the CF label is redundant. i disabled it but it still showed up.

Osiris-DevWorks:
this is because I didn't click "apply tags", so workflow needs to be clearer, and perhaps generate enhancements should have an apply tags hook.

Issue #172, opened 2026-06-20
Persist data directory across installs, keep it out of OneDrive, and back up user.ini


Osiris-DevWorks:
## Summary

Smart Citizen needs to reliably remember the user-chosen data directory across reinstalls and upgrades, stop defaulting users into a OneDrive-synced folder, and keep a backup of `user.ini` so a lost override file is a one-click restore instead of a recovery dig.

This comes out of a real incident: a user's `user.ini` (36 favorited ships plus a few commodity edits) was emptied to 0 bytes. The data had been sitting under a Documents folder that Windows had redirected into OneDrive, and OneDrive churned it: by the time we looked, the same `user.ini` existed as a 0-byte placeholder in three places (`Documents\Smart Citizen`, `OneDrive\Documents\Smart Citizen`, and a stray `OneDrive\Pictures\LIVE` the "back up my Pictures" feature had swept up). The data root had silently moved three times across those locations over a few weeks. Recovery was only possible because the applied game `global.ini` still carried the favorites and an old backup happened to exist on disk.

### 1. Persist the data directory across installs (primary)

The user sets a data folder once. It should survive uninstall, reinstall, and version upgrades without reverting. Today the location is not sticky in practice (it moved across three folders), and the installer wizard's data-folder page can quietly fall back to the default if the user doesn't re-pick it every time.

- Audit the full round-trip: `WriteInstallerChoicesToRegistry` and the data-folder wizard page in `installer.iss`, against `AppSettings.get_user_data_dir()` / `set_user_data_dir()`.
- The wizard should pre-fill the previously chosen folder and re-persist it on every install, never silently revert to Documents.
- Confirm uninstall never drops `user_data_dir` (the dead `CleanRegistrySettings()` targets only the legacy node and is never called today, but lock this with intent so a future change can't reintroduce the wipe).

### 2. Stop defaulting users into OneDrive

The default data root is `Documents\Smart Citizen`. When Windows has redirected Documents into OneDrive (very common), the app's per-user files land inside OneDrive's reach, where Files On-Demand and folder-backup churn can dehydrate or empty them.

- At startup (and on the installer's data-folder page), detect when the resolved data root sits under a OneDrive-managed path and warn, with a one-click "move my data somewhere local" option.
- Detection: compare the resolved root against `%OneDrive%` / `%OneDriveConsumer%` and the `User Shell Folders\Personal` target.

### 3. Back up `user.ini`

[...]

PR #311, opened 2026-07-25
Add Export / Import Settings backup

Stealrull:
## Why

Portable users re-do their whole setup on every update — each release unzips into a fresh versioned folder with an empty `data/`, stranding the previous one. This is also the "I got a new PC" path.

Adds a **Settings Backup** group to the Config tab: **Export Settings…** / **Import Settings…**, producing one small zip (a few KB in practice).

## What's in the backup

| Included | Excluded |
|---|---|
| Every backend setting (preferences, tag configs, category toggles) | DataForge / `base.ini` cache — tens of MB, rebuilds from the game |
| The Star Citizen install path | `backups/` |
| Each channel's `user.ini` | Data-root and cache-dir overrides |

The data-root/cache overrides are deliberately excluded: pointing them at a drive that doesn't exist on the target machine would strand every path derived from them. Measured on a real profile, the cache was 14 MB while the actual settings were under 1 KB.

## Cross-version by design

The zip carries a manifest (schema version, exporting app version, source mode, channels). A backup made by any build imports into any other; import only refuses a schema **newer** than the running build understands. The filename omits the app version so it doesn't imply otherwise — the exporting version is still recorded in the manifest and shown in the import confirmation.

## Import is validated, and reversible

1. Validates the zip (rejects non-backups, unreadable manifests, newer schemas) and shows what's inside before doing anything.
2. Snapshots current `user.ini` files into the channel's rotating backups (#172 machinery) — so an unwanted import is undoable via **Restore user.ini**.
3. Reconciles the Star Citizen path: **kept** when the folder actually exists here, **dropped in favour of auto-detection** when it doesn't — a backup from another machine can't pin the user to a dead path.
4. Sets a one-shot flag and offers a restart. The next launch offers to regenerate + apply enhancements, reusing the Simple-mode generate-then-apply continuation (#180).

Restart is required because most settings are read once into widgets at construction; there's no existing full-UI-rehydrate path (same limitation `i18n.py` documents for language switching).

## Two lifecycle details worth reviewing

- **Export flushes on-screen Tag Builder edits first**, so the backup captures what the user is looking at rather than the last-saved state — the same "what you see is what you save" rule Generate Enhancements follows (#215).
- **Import refreshes the Tag Builder pages and category checkboxes afterwards.** These widgets cache settings at construction, so an import that rewrote the `tag_builder/*` keys underneath a live tab left the pages holding the pre-import config — and the next Save Tag Changes / Generate / Export wrote that stale state straight back over the import. This was a real, reproduced bug found in testing. `_reset_to_defaults` is refactored into `apply_config()` so the refresh reuses the existing widget-resync logic rather than duplicating it; Reset keeps persisting General Tags defaults, the import path only mirrors them.

## Both build variants

Export/import go through the `AppSettings` backend abstraction, so this works identically in registry and portable builds — one zip format moves a setup between the installer and portable variants in either direction.

## Dependency note

Import calls `get_sc_install_root()`, so it inherits whatever detection that method offers. Full multi-drive auto-detection (#269) reaches this branch with the `release/2.2.1` merge in #289; until that lands, the fallback here is the two hardcoded Program Files paths. No code change needed either way — this picks it up automatically.

## Testing

- `tests/test_settings_profile.py` — round trip, manifest validation, machine-key filtering, install-path reconciliation, cross-version import, crafted-zip path traversal.
- `tests/test_tag_builder_import_refresh.py` — the import-refresh fix, including a test that locks the **failure mode itself** so a future refactor can't silently undo it.
- Full suite: **1238 passed, 0 failed**.
- Manually verified on a frozen portable build: customise → export → wipe `data/` → import → restart → settings, game folder, and Tag Builder config all restored; confirmed the previously-broken "Restart Later → Generate Enhancements" path no longer clobbers.

### DIFF
--- a/docs/HELP.md
+++ b/docs/HELP.md
@@ -98,7 +98,7 @@ Use **Reset user.ini** in the **Config** tab to wipe all of your personal edits 
 
 ## 13. Export / Import Settings
 
-Use **Export Settings** and **Import Settings** in the **Config** tab to move your whole Smart Citizen setup between PCs, or back it up before a fresh install. Export bundles your app settings and every channel's `user.ini` overrides into one small zip; secrets and machine-specific paths (your Star Citizen install path, window geometry) are left out. Import layers that backup over your current settings and replaces `user.ini` for the channels it contains — your current `user.ini` files are snapshotted first via **Restore user.ini**, so an import is reversible. Smart Citizen restarts after an import to load the new settings, then offers to regenerate and apply your enhancements.
+Use **Export Settings** and **Import Settings** in the **Config** tab to move your whole Smart Citizen setup between PCs, or back it up before a fresh install. Export bundles your app settings and every channel's `user.ini` overrides into one small zip, including your Star Citizen install path; machine-specific paths that wouldn't make sense on another PC (your data folder, cache location, window geometry) are left out. Import layers that backup over your current settings and replaces `user.ini` for the channels it contains — your current `user.ini` files are snapshotted first via **Restore user.ini**, so an import is reversible. Your Star Citizen path is only kept if it still exists on the PC you're importing to; otherwise Smart Citizen auto-detects it instead. Smart Citizen restarts after an import to load the new settings, then offers to regenerate and apply your enhancements.
 
 ## 14. After Game Updates
 
--- a/docs/HELP.md
+++ b/docs/HELP.md
@@ -96,11 +96,15 @@ Open the **More** menu and choose **Export INI…** to bundle the currently appl
 
 Use **Reset user.ini** in the **Config** tab to wipe all of your personal edits for the active channel. A confirmation prompt makes sure it isn't a mis-click, and an automatic backup of the current `user.ini` is taken into `<data folder>\<channel>\backups\` first — so a reset is recoverable if you change your mind.
 
-## 13. After Game Updates
+## 13. Export / Import Settings
+
+Use **Export Settings** and **Import Settings** in the **Config** tab to move your whole Smart Citizen setup between PCs, or back it up before a fresh install. Export bundles your app settings and every channel's `user.ini` overrides into one small zip; secrets and machine-specific paths (your Star Citizen install path, window geometry) are left out. Import layers that backup over your current settings and replaces `user.ini` for the channels it contains — your current `user.ini` files are snapshotted first via **Restore user.ini**, so an import is reversible. Smart Citizen restarts after an import to load the new settings, then offers to regenerate and apply your enhancements.
+
+## 14. After Game Updates
 
 When Star Citizen updates, your edits are preserved in `<data folder>\<channel>\user.ini`. Re-run **Extract from Data.p4k** to pull fresh stock strings from the patched game — the table reloads automatically and your customizations re-apply on top.
 
-## 14. Switch Languages
+## 15. Switch Languages
 
 Pick a language from the **Language** dropdown in the **Config** tab (next to Channel). Switching changes both the app's interface and the game strings in the table:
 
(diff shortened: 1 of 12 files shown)


Osiris-DevWorks (changes requested):
Well-structured feature (Qt-free pack/unpack, routes through `AppSettings` so registry/portable parity holds, machine-specific keys filtered on both export and import, good test coverage). One blocking security gap plus minors.

**Blocking — path traversal on import (`src/utils/settings_profile.py` → `read_profile_zip`)**
The channel-segment guard only rejects a forward slash: `if not channel or "/" in channel: continue`. A channel of `..` or a backslash segment (`..\evil`) passes it. `_handle_import_settings` then does `target = get_channel_user_ini_path(channel)` → `user_data_dir\..\user.ini`, `target.parent.mkdir(parents=True, exist_ok=True)`, `target.write_text(text)` — on Windows this writes attacker-controlled INI content **outside the data root**. The PR states traversal protection, but `test_nested_override_entry_skipped` only exercises the `/` case. Fix: reject any channel that is `.`/`..` or contains a backslash / `os.sep` / `os.altsep`, or restrict the import loop to `AppSettings.AVAILABLE_CHANNELS` (do the strict segment check in `settings_profile`, which is settings-free, for defense in depth). Low real-world likelihood since users import their own backups, but it's a stated guarantee and trivially closeable.

**Minor**
- `src/utils/settings.py` `export_all_values` — `import json as _json` is redundant; `json` is imported at module top.
- Docs drift: the new Export/Import Settings feature isn't mentioned in `docs/HELP.md`.

**Rebase**: now conflicts on `languages/english/ui.json` against the merged i18n work (#324) — rebase onto the current tip.

---

Stealrull (commented):
## Self-review

Reviewed all 12 files (utils → gui → tests → docs → i18n).

**Findings, both confirmed with you and fixed in `647489e`:**
- `docs/HELP.md`'s new "Export / Import Settings" section claimed the Star Citizen install path is left out of the backup. It isn't — `settings.py` deliberately keeps it (`sc_install_root`/`game_install_path` are explicitly excluded from `PROFILE_EXCLUDE_KEYS`), validated on import via `reconcile_imported_install_path()` with a dedicated test (`test_sc_install_path_travels`), and the in-app UI strings already say so correctly. Only the doc paragraph was wrong; fixed to match the actual (well-reasoned) behavior.
- `PROFILE_EXCLUDE_KEYS` hardcoded the literal `"post_import/apply_pending"` instead of referencing `POST_IMPORT_APPLY_PENDING`, defined nine lines above in the same class body. No test cross-checks the two stay in sync (the existing test parametrizes over the frozenset itself), so a future rename would silently desync them. Replaced the literal with the constant reference.

**Verified clean:**
- The path-traversal fix (`settings_profile.py`, addressing the earlier blocking review) is thorough: rejects empty, `.`, `..`, `/`, `\`, `os.sep`, and `os.altsep` in channel names, with dedicated regression tests for each variant.
- The Tag Builder import-refresh (`enhancements_tab.py`) correctly resyncs on-screen state after an import so Save/Generate/Export can't clobber it with stale pre-import config — has a regression test that deliberately locks the *bug's own failure mode* (`test_stale_pages_clobber_import_without_refresh`), not just the fix.
- `main_window.py`'s restart flow (`_suppress_user_ini_autosave`) correctly gates both the unapplied-edits nag and the autosave-clobber path, only set right before a post-import restart, never otherwise.
- The `_reset_to_defaults` → `apply_config()` refactor preserves reset-vs-import semantics correctly (own dedicated test: `test_reset_to_defaults_still_persists_title_tags`).
- Test coverage is thorough: 672 lines across the two new test files covering round-trip, validation, traversal, exclusions, both backend shapes, install-path reconciliation, and the export-flush-before-read race.
- Non-blocking note: `read_profile_zip` decompresses zip entries fully into memory with no size cap — low priority since this is a user importing their own local file, not adversarial internet input.
- French/Spanish/Portuguese settings-backup translations (added in an earlier commit on this PR) checked out against the English source.

Suite: 1391 passed, 21 skipped, same 5 pre-existing Chinese/Italian/Japanese activation failures tracked by #318, unrelated to this PR.

PR #320, opened 2026-07-26
[#319] Drop the "+" from mission-description reputation reward text

Stealrull:
Part of #319 (the mission-details "+" half; the main-menu watermark line-break half is #304).

## Problem

The MISSION DETAILS body prepended a "+" to the reputation reward amount ("Rep: +500"), which read as confusing rather than informative sitting next to a plain number. There's no companion "-500" a reader would be contrasting it against on the success line, so the sign carried no information.

## Fix

Dropped the "+" from all three call sites that built it: the pu_missions/entities single-mission path, and both the single-tier and multi-tier contractgen paths. A failure-penalty amount is untouched, that one's own "-" sign stays since it genuinely does need to read as negative next to the reward line above it.

`_rep_reward_line()` itself is unchanged: it has always been a dumb passthrough for a pre-formatted `amount_str`, so the sign was always the caller's decision, not something baked into the formatter.

## Testing

Two new cases in `tests/test_mission_rep_label.py` exercising `enhancements_mission()` end-to-end (not just the passthrough formatter) to lock down that a reward gain renders unsigned while a penalty keeps its "-". Full suite green: 1189 passed, 16 skipped on this branch. Built a portable combined with every other in-flight 2.3.0 branch and confirmed in-game.

### DIFF
--- a/scripts/generate_enhancements_ini.py
+++ b/scripts/generate_enhancements_ini.py
@@ -2753,13 +2753,15 @@ def _rep_reward_line(field_name: str, amount_str: str, rep_xp_label: str, track:
     """Render one reputation-reward line for a MISSION DETAILS body.
 
     *field_name* is the rank / standing name when known (e.g. ``"Neutral"``),
-    otherwise ``""``. *amount_str* is the pre-formatted signed amount, e.g.
-    ``"+500"``, ``"+500–4,000"``, or ``"-100"``. The configured reputation
-    label (``rep_xp_label``, default ``"Rep"``) becomes the field name when no
-    rank is known, and the trailing unit otherwise; it is never doubled. The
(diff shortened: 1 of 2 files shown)


Osiris-DevWorks (approved):
Approving. Correct and complete for its scope.

- All three rep-reward body call sites drop the `+`; the failure-penalty `-` sign and unrelated `+` usages (overflow tag, title `[… Rep]`) are correctly left alone.
- `_rep_reward_line()` stays a passthrough — the sign was always the caller's decision.
- Two new end-to-end cases in `tests/test_mission_rep_label.py` lock reward-unsigned / penalty-keeps-minus.

Satisfies the mission-details half of #319 (the watermark half is #304).

---

Issue #318, opened 2026-07-26
i18n: the 4 new languages need a 253-key backfill once the 2.3.0 PRs are merged


Stealrull:
## Summary

Merging all the in-flight 2.3.0 work together leaves German, Chinese, Italian and Japanese each missing **253 UI keys**, which fails 8 tests. Nothing is wrong with any individual PR — this only appears once they're combined, which is why nobody has seen it (see #317: CI doesn't run on `release/2.3.0`).

## Why it happens

Each language PR is exactly self-consistent on its own branch:

| PR | Language | English keys on branch | Translated keys | Missing | Extra |
|---|---|---|---|---|---|
| #313 | German | 376 | 376 | 0 | 0 |
| #314 | Chinese | 376 | 376 | 0 | 0 |
| #315 | Italian | 376 | 376 | 0 | 0 |
| #309 | Japanese | 376 | 376 | 0 | 0 |

Those branches were cut before #268, #308, #311 and others added new English keys. Merged together the English file reaches **629** keys, so each language ends up 253 short — and every `test_*_activation.py` asserts full coverage with every leaf `at`-filled.

Note this **cannot be fixed inside the language PRs**: adding the keys now would make each language file contain keys its own branch's English file doesn't have, tripping the same tests' "no extra keys" assertion. It has to land after the merge.

## What's missing

The 253 keys per language are, roughly: the `settings_backup.*` dialogs and `config.*` export/import strings (#311), the Blueprint Tracker `force_rescan_*` (#308) and `scan_other_channels_*` (#268) checkbox strings, the mission-detail-field and Tag Builder mapping tooltips, and assorted `status_bar.*` / `progress.*` / `test_plan.*` / `apply.*` strings.

## A second post-merge landmine: `import_dialog.custom_value_prompt`

The merged `tr()` kwarg-collision fix ("tr() kwarg named 'key' collides with its own first parameter") renamed this string's placeholder `{key}` -> `{loc_key}`, and `import_dialog.py` now calls `tr("import_dialog.custom_value_prompt", loc_key=key)`.

That rename is on `release/2.2.1` and reaches 2.3.0 only through #289. All four new language files still carry the old `{key}`:

- `release/2.3.0` English today: `Enter custom value for key:\n{key}`
- `release/2.2.1` English (arrives via #289): `Enter custom value for key:\n{loc_key}`
- all four new languages: `...\n{key}`

[...]

PR #326, opened 2026-07-30
[#325] Make commodities Tag Builder elements default to off

### LINKED ISSUE #325: Make commodities tag builder tab options default to off
Users have reported problems with commodities tab (Label, Used to Craft, Collection) tagging in game, having them default too off and only be turned on by the user allows them to control if they run into an issue/unknown text.

Stealrull:
## Summary
- Closes #325.
- Users reported garbled/unknown text from commodity tagging showing up in-game.
- The commodities Tag Builder config's three elements (Label/CF, Usage/"Used To Craft", Collection) all shipped enabled by default, so every fresh install tagged commodity names without the user ever choosing to.
- Fresh installs now show plain commodity names until the user opts in per element from the Tag Builder. Existing users' saved configs are untouched — `get_tag_config` only falls back to the default when nothing is saved yet.
- `_backfill_new_elements` also reads its enabled state from the same default table, so a commodities config missing one of these three elements (e.g. a pre-2.1 config predating "usage") now backfills it disabled too, instead of picking up a newly-added element enabled without the user opting in.

## Testing Checklist
- [x] PR is scoped to one concern (default-value change, no new feature)
- [x] `pytest tests/` passes locally (1447 passed, 21 skipped, same 5 pre-existing Chinese/Italian/Japanese failures unrelated to this change)
- [x] User-facing strings, `docs/HELP.md` reviewed for drift (updated the Tag Builder section to list all three commodity elements and note they're disabled by default, matching the existing Components Type element note)
- [x] New/changed behavior has matching test coverage (updated tests that relied on the old always-on default to build their own explicitly-enabled config where they test render logic, and flipped the two tests that asserted "on by default" to assert the new default)

## Testing performed
Ran the full test suite, built a portable, and had it tester-confirmed working before opening this PR.

---


Stealrull (commented):
Reviewed this file-by-file since it's my own PR from earlier today. All three commodities Tag Builder elements (Label, Used To Craft, Collection) now default off, matching what #325 asked for; existing users' saved configs are untouched since `get_tag_config` only falls back to the default when nothing is saved yet.

Self-review found and fixed two issues, now pushed:
- `_backfill_new_elements`'s docstring in `src/utils/settings.py` overclaimed that the enabled-by-default state was "disabled for every category as of #325" — checked the actual `DEFAULT_TAG_CONFIGS` data and that's wrong: only commodities' three elements and components' `type` default off, missiles/ship_weapons/mission_titles/components' class/size/grade all still default enabled. Corrected.
- An em dash I introduced while editing the `docs/HELP.md` Tag Builder line (the line's other two em dashes predate the no-em-dash convention and were left alone).
- `tests/test_commodity_journal.py`'s new helper landed between the import block and `pytestmark`, breaking the file's established ordering. Reorganized.

Full test suite passes (same 5 pre-existing Chinese/Italian/Japanese failures, unrelated, tracked by #318).

Approving.

Osiris-DevWorks (approved):
Verified the change is scoped to the three commodities elements in DEFAULT_TAG_CONFIGS, saved user configs are untouched, the backfill inherits the disabled default with a test locking it, and the generator call sites guard the empty-tag case. Test updates are coherent flips, not deletions. Merging.

PR #328, opened 2026-07-30
[#318] Post-merge i18n backfill for Chinese, Italian, Japanese

Stealrull:
## Summary
- Closes #318.
- Each of the four new-language PRs is individually self-consistent on its own branch, but those branches were cut before #268, #308, #311 and others added new English UI keys. Once merged, each language ends up 215 keys short of English's key universe, and each language's own `test_*_activation.py` asserts full coverage with every leaf `at`-filled -- so the merged result fails tests that no individual PR fails.
- German already got this treatment during its own PR's rebase (#313, not yet merged as of this PR). This backfills the three languages already merged onto `release/2.3.0`: Chinese, Italian, Japanese.
- Backfilled 215 keys per language (645 strings total) styled on each file's existing terminology, with placeholders, leading/trailing whitespace, and untranslated literals (`user.ini`, `Data.p4k`, `DataForge`, `LIVE`/`PTU`/`EPTU`/`HOTFIX`, `[Owned]`, `[BP]`) preserved.
- Fixed the second post-merge landmine #318 flagged: `import_dialog.custom_value_prompt`'s `{key}` placeholder was renamed to `{loc_key}` on `release/2.2.1` and reached 2.3.0 via #289, but all three languages still carried the old `{key}` -- not wrong on their own branches (English said `{key}` there too), but a live `KeyError` once merged, since `import_dialog.py` calls `tr(..., loc_key=key)`. Fixed all three to `{loc_key}`.
- Fixed the `{plural}` inconsistency #318 flagged: `dialogs.suppressed_errors_suffix`'s `{plural}` token is only ever filled with a hardcoded English "s" suffix (`main_window.py`), which has no correct value in Chinese, Italian, or Japanese. German already drops this token via an `_ALLOWED_PLACEHOLDER_DROPS` allowance; added the same allowance to the other three languages' activation tests and dropped the token from their translations, matching `TRANSLATIONS.md`'s documented policy.

## Testing Checklist
- [x] PR is scoped to one concern (post-merge i18n backfill for 3 languages, no app code changed)
- [x] `pytest tests/` passes locally (1335 passed, 0 failed, 21 skipped -- the 5 pre-existing failures this backfill targets are gone)
- [x] Portable built and tester-confirmed working
- [x] Target branch is the active `release/2.3.0`, not `main`

## Testing performed
Ran the full test suite before and after (5 failed -> 0 failed). Built a portable merging this alongside the other in-flight fixes and had it tester-confirmed working before opening this PR.

---
[...]

PR #313, opened 2026-07-25
Add German language support

Stealrull:
Closes #299.

## Summary
- Full AI translation of all 376 UI keys plus the 19-step guided tour (`tutorial.*`), all `at`-only (`ht` empty — no human translator yet)
- Translated `HELP.md`, `ABOUT.md`, `LEGAL.md`, and `FAQ.md`
- Base `global.ini` sourced from [rjcncpt/StarCitizen-Deutsch-INI](https://github.com/rjcncpt/StarCitizen-Deutsch-INI) (`live/global.ini`, the standard hybrid translation, not the extended "Deutsch+" variant)
- `SC_LANGUAGE_IDS["german"] = "german_(germany)"`; installer `LanguageChoicePage` gained a German option as the 5th language
- Locked by `tests/test_german_activation.py` (mirrors `test_japanese_activation.py`)

Also fixes a pre-existing bug affecting every language: switching language while the app is running never re-rendered the About/FAQ/Legal tab bodies — only the toolbar/tabs/table — so a restart was silently required. `retranslate_ui()` now re-renders all three.

## Test plan
- [x] Full test suite passes (`pytest tests/`)
- [x] Portable build tested manually: German selectable in Config tab, UI/tutorial/docs render in German, switching to/from German mid-session updates About/FAQ/Legal without a restart
- [ ] Human review of the AI translation (tracked in `languages/TRANSLATIONS.md`, grep `"ht": ""`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #330, opened 2026-07-31
[#329] Restrict ship favoriting to name rows, add Ship/Vehicle Names Only filter

### LINKED ISSUE #329: Ship descriptions can currently be favourited which shouldn't be possible, suggestion to make a "Vehicle Name Only" tickbox for string editor
Ship descriptions can be favourited currently, this is an un needed feature and can confuse players when selecting "ships" from the dropdown box.

Suggestion to make a "Ship/Vehicle Name Only" tickbox before Favourites Only so people can easily select favourites for ASOP terminals.

Image

Stealrull:
## Summary
- Closes #329.
- **Ship descriptions could be favorited, which shouldn't be possible.** Description entries (`vehicle_Desc*`, Wikelo `*_VehicleDesc`) share the "Ships" category with name entries, but the favorite-prefix / sort-order mechanism (`ship_sort_prefix.py`) only ever reads the NAME entry's `custom_value` — that's what feeds the in-game ASOP terminal display name. Favoriting a description did nothing in-game but confused the Favorites Only filter and the Ships dropdown.
- Added `is_ship_name_key()` to `string_model.py` and a paired `is_favoritable_ship(entry)` helper, then used the helper everywhere the star/sort-order mechanism touches an entry: display, editability (`flags()`), tooltips, foreground/background colors, sort keys, `COL_ORDER`'s `setData`, `toggle_favorite()`, the star-click handler, and the context menu's Add/Remove Favorite action. Description rows now show a blank star/order column and can't be favorited by any route.

## New filter: Ship/Vehicle Names Only
Added the requested checkbox, placed before Favorites Only since the two are meant to be used together when picking ASOP favorites. It narrows the table to **only** ship/vehicle name rows — hiding ship descriptions *and* every other category — leaving exactly the rows favorites and ASOP sort order apply to.

The label uses a slash rather than an ampersand deliberately: Qt reads `&` in a widget label as a mnemonic marker, so "Ship & Vehicle Names Only" rendered as "Ship <ins>V</ins>ehicle Names Only" with the ampersand swallowed. There's a comment at the widget noting this so it doesn't get "fixed" back.

## Heads-up: Favorites Only semantics changed slightly
Worth flagging since it goes marginally beyond what #329 literally asked for. **Before:** any row whose `custom_value` started with the favorite prefix passed the Favorites Only filter, regardless of category. **After:** the row must also be a ship/vehicle name row.

This looks like a latent false positive rather than a feature. A non-Ships row could never actually *be* favorited — the star column never rendered for it, and neither the star-click handler nor the context menu would toggle it (both already gated on `category == "Ships"` before this PR). So any such match was a coincidental prefix collision, e.g. a Gear row a user happened to start with `*`. The new behavior also makes the filter agree with what the star column actually displays. Documented in `filter_entry_indices`' docstring; happy to revert this half if you'd rather keep the old behavior.

## Note on `*_VehicleNameShort`
Short-name variants deliberately still count as favoritable. Base-game `vehicle_Name*_short` keys never reach a `StringEntry` (they're dropped in `ini_parser`'s build loop), so this only admits Wikelo's `*_VehicleNameShort` rows — which were already favoritable before this PR, since they carry the Ships category. #329 was about descriptions only, so their behavior is deliberately unchanged. Documented in the helper's docstring so it reads as a decision rather than an oversight.

## Testing Checklist
- [x] PR is scoped to one concern (#329's two linked asks)
- [x] `pytest tests/` passes locally (1349 passed, 21 skipped; the 5 failures are the pre-existing Chinese/Italian/Japanese key-coverage gap tracked by #318 and fixed separately in #328)
- [x] User-facing strings, `docs/HELP.md` reviewed for drift (Ship Favorites section updated for both the name-row restriction and the new checkbox)
- [x] New logic has matching test coverage (`tests/test_ship_name_desc_split.py` for both helpers, plus new cases in `tests/test_entry_filter.py`)
- [x] DRY: the two-part predicate is stated once in `is_favoritable_ship`, not inlined at 18 call sites

## Why `is_favoritable_ship` keeps both halves
It'd be tempting to simplify it to a bare `is_ship_name_key(entry.key)`, since every name-shaped key *usually* also has category "Ships". Don't: `StringEntry.category` isn't always derived from the key. `ini_parser` assigns it from the source enhancement file when `enhancements_key_categories` applies, so the stored category is the authority on which bucket a row displays under. There's a regression test pinning that exact case so a future simplification pass gets a red test rather than a silent behavior change.

## i18n
Translated into French, Spanish, Portuguese (Brazilian), Chinese, Italian, and Japanese alongside English, to avoid opening another instance of the post-merge key-coverage gap #318 tracks. German's own PR (#313) isn't merged upstream yet, so this branch has no German file to update — that key will need adding whenever #313's branch is next touched.

## Testing performed
Ran the full suite, built a portable merging this alongside the other in-flight 2.3.0 work, and had it tester-confirmed working (including two rounds of corrections after testing: the filter originally only hid ship descriptions and left other categories visible, and the label originally used an ampersand).

---


Stealrull (commented):
## Self-review

Reviewed layer by layer (models → utils → gui → tests → docs → i18n) against #329's two asks: stop descriptions from being favoritable, add a "Ship/Vehicle Names Only" filter scoped to Ships-category name rows only.

**Findings, fixed in `b8ed698`:**
- The `entry.category == "Ships" and is_ship_name_key(entry.key)` predicate was repeated 18 times across `string_table_model.py`, `entry_filter.py`, and `main_window.py` — every single `category == "Ships"` check in those three files turned out to be part of it. Extracted to one named helper, `is_favoritable_ship(entry)`.
- Kept both halves of the predicate rather than collapsing to a bare `is_ship_name_key(entry.key)`: `StringEntry.category` is not always key-derived (`ini_parser` can override it via `enhancements_key_categories`), so the stored category is the real authority on which bucket a row displays under. Added a regression test pinning that exact case so a future simplification pass can't silently drop it.
- Documented a real (intentional, not a regression) behavior change: `favorites_only` used to pass any row whose `custom_value` merely started with the favorite prefix, regardless of category. Now it requires `is_favoritable_ship` too. This is a latent-bug fix — a non-Ships row could never actually be favorited (the star column never rendered for it, neither the click handler nor the context menu would toggle it) — but it wasn't called out in the original PR description, so I added it.
- Documented the `*_VehicleNameShort` decision in `is_ship_name_key`'s docstring (stays favoritable, matches pre-#329 behavior, #329 was about descriptions only) so it reads as deliberate rather than an oversight.

**Independently verified via adversarial multi-agent review:** logical equivalence of every inlined-predicate→helper call site (including the negated forms in `flags()`/`data()`), no missed or over-eager call sites, no import cycles, and empirical behavior-equivalence across 91 key/category combinations comparing the old inline predicate against the new helper. All four lenses came back clean.

**Verified clean otherwise:**
- `is_ship_name_key()` correctly excludes base-game `vehicle_Name*_short` entries (never reach a `StringEntry` — dropped in `ini_parser`'s build loop before parsing).
- Star/order column gating (display, editability, tooltips, colors, sort keys) and the favorite-toggle guard all route through the same helper — no path left using the old inline check.
- Qt mnemonic bug avoided deliberately in the checkbox label ("Ship/Vehicle Names Only" uses a slash, not `&`), with a code comment explaining why so it doesn't get "fixed" back to an ampersand.
- i18n: 7 languages translated on this branch; German is local-integration-only since its own PR (#313) isn't merged upstream yet, documented in the PR body.

Suite: 1349 passed (feature branch state at time of review), 21 skipped, same pre-existing Chinese/Italian/Japanese activation failures tracked by #318.

PR #332, opened 2026-07-31
[#331] Add independent RS ore-name annotation and mission-details breakdown toggles

### LINKED ISSUE #331: Show RS number next to ore name in mission text for battaglia missions
While we added RS to mission titles I avoided adding it to mission text, however this turns out to be useful as it's that text that shows in the top right mission tracker and users have asked for it.

Will have it listed as it's own tickbox in Localization Enhancements and mission details field

Stealrull:
## Summary
- Closes #331.
- Battaglia scan/mining contracts already carry a flat `[RS ####]` tag on the mission title, but the value never showed up anywhere a player actually reads the contract while working it. This restores work from an earlier, unmerged attempt (commits `6891890`/`739b6f8` on the abandoned `release/2.3.0-local` branch) that put the value where players actually read it, now split into the two independent checkboxes the issue asks for instead of one bundled toggle.

## Localization Enhancements: "Show Resource Signatures (RS) next to ore names"
On by default. Appends each mineable ore's base RS to its own display-name loc key (`mineabletype_primary_<ore>`), so it shows up everywhere the game renders that name, including the top-right mission tracker, which is the specific case users asked for.

This is the half that was pulled entirely from the earlier attempt for being "too broad an effect... for what it bought" (it touches every place the game shows an ore's name, not just Battaglia missions). Reviving it as its own opt-in toggle addresses that concern directly instead of re-bundling it.

## Mission Detail Fields: "Resource Signatures"
Also on by default. Adds a breakdown to the mission's own DETAILS body listing each targeted ore's full RS value progression (curated per-ore stack values, distinct from the title tag's single flat number). Independent of the ore-name checkbox above and of the existing mission-title `[RS ####]` General Tags toggle.

## Also carries forward a real bug fix
`_build_battaglia_mineable_rs_tags` only read the inline Description `ContractStringParam`, so any contract inheriting its desc from a shared body template (the `"template"` UUID fallback `scan_contract_generators` already handles) silently got no desc-level RS data at all, even though its title tag worked fine.

## Testing Checklist
- [x] PR is scoped to one concern (#331)
- [x] `pytest tests/` passes locally (1500 passed, 21 skipped; the languages on this branch are fully covered — Chinese/Italian/Japanese have their own pre-existing key-coverage gap against `upstream/release/2.3.0` tracked by #318, unrelated to this branch)
- [x] User-facing strings, `docs/HELP.md` / `docs/ABOUT.md` reviewed for drift
- [x] New logic has matching test coverage (`tests/test_rs_ore_name_annotations.py`, extended `tests/test_battaglia_mineable_rs.py` and `tests/test_mission_detail_fields.py`)

## i18n
Translated into French, Spanish, and Portuguese (Brazilian) alongside English. Chinese, Italian, and Japanese already have a pre-existing key-coverage gap against `upstream/release/2.3.0` from before this branch was cut (the recurring post-merge backfill pattern tracked by #318/#328); adding 3 more keys there wouldn't change whether their activation tests pass, so it's left for that follow-up pattern rather than done piecemeal here. German has no file on this branch at all since its own PR (#313) isn't merged upstream yet.

## Testing performed
Ran the full suite, built a portable merging this alongside the other in-flight 2.3.0 work, and had it tester-confirmed working. Iterated twice after initial testing: the ore-name checkbox was renamed from "Show RS in ore names" to "Show Resource Signatures (RS) next to ore names" for clarity, and both checkboxes were flipped from off-by-default to on-by-default per tester feedback.

---


Stealrull (commented):
## Self-review

Reviewed layer by layer (settings → generator → GUI wiring → tests → docs/i18n) against the PR's own stated goal: two fully independent toggles, both on by default, matching the earlier-established pattern for the sibling fields around them.

**Findings, both fixed in `9080076`:**
- `EnhancementsGeneratorWorker.__init__`'s `rs_ore_name_annotations` parameter still defaulted to `False`, left over from before the default-on flip. Never actually hit in the running app (`main_window.py` always passes an explicit value from `AppSettings.get_rs_ore_name_annotations()`), but it disagreed with `AppSettings`' and `generate_enhancements_ini.main()`'s defaults, which are both `True`. A direct instantiation without the kwarg (a future caller, a test) would have silently gotten the annotation off while everything else says on.
- Two code comments (in `enhancements_tab.py` and `generate_enhancements_ini.py`) still called the checkbox "Show RS in Ore Names" — its name before the rename to "Show Resource Signatures (RS) next to ore names". Also reworded one comment's "opt-in toggle" phrasing, which read as off-by-default and no longer matched the code.

**Verified clean:**
- Both settings' defaults (`AppSettings.get_rs_ore_name_annotations`, `get_mission_detail_fields()["resource_signatures"]`) are `True`, consistent with `main()`'s kwarg defaults and the ctx.get fallbacks inside `_run_gen_missions`.
- `_MISSION_FIELD_DEFAULTS` was removed rather than left as an empty override dict once nothing needed a non-True default — no unused infrastructure left behind.
- The two toggles are genuinely independent: different settings keys, different code paths (`_build_mineable_rs_name_overrides` patches a loc key at generation end; `_format_rs_details_lines` gates a DETAILS body line), confirmed by `test_resource_signatures_mission_field_is_independent`.
- The template-fallback bug fix in `_build_battaglia_mineable_rs_tags` (contracts inheriting `Description` from a shared template) is covered by dedicated regression tests (`test_desc_key_resolved_via_template_when_no_inline_desc_param`, `test_inline_desc_param_wins_over_template`).
- i18n: French/Spanish/Portuguese fully translated and consistent with the established RS-abbreviation terminology already used by `mt_tag_rs_cb`. Chinese/Italian/Japanese activation-test failures are pre-existing (confirmed present on `upstream/release/2.3.0` before this branch touched anything) and tracked separately by #318.
- Docs (`HELP.md`, `ABOUT.md`) and the test plan accurately describe both toggles as on by default.

Suite: 1500 passed (integration branch, includes both this PR and everything else currently merged locally), 21 skipped, 0 failures.

Osiris-DevWorks (approved):
Verified the toggle independence (separate keys and code paths, both tested), the worker default fix in 9080076, the _MISSION_FIELD_SETTING pattern for resource_signatures, and the template-fallback regression tests. Two stale comment overstatements about the (RS) annotation depending on desc resolution can be tidied in a follow-up; behavior and tests are correct. Merging.

PR #335, opened 2026-08-01
Fix raw blueprint-filename bullets not resolving to display names

Stealrull:
## Problem

Some mission text (CIG's own `UWC_Refueling` bodies) ships the raw blueprint XML filename stem verbatim in a POTENTIAL BLUEPRINTS bullet, e.g.:

```
bp_craft_nozzle_fuelgiver_grin_nozzleverysecure
```

instead of a resolved display name. The Blueprint Tracker then lists that raw string as its own item, so the real item ("Harkin") never gets an `[Owned]` tag from that bullet.

This is a different shape than the existing `_key_slug` garbled-KEY bug, so it never matched `_BULLET_NAME_ALIASES` even though the alias table already had the right answer.

## Fix

Adds `_raw_blueprint_filename_slug()`, which mirrors the prefix-strip + title-case transform `scripts/generate_enhancements_ini.py`'s `_name_from_blueprint_filename` already performs, then re-checks `_BULLET_NAME_ALIASES`. It runs as a fallback in the existing bullet-resolution chain, only after the current alias and `_key_slug` attempts have both failed, so no bullet that resolves correctly today changes behaviour.

Prefixes handled: `bp_craft_`, `bp_rewards_`, `bp_`.

## Testing

Confirmed against the real `UWC_Refueling` mission body in `tests/fixtures/kraken_global_latest.ini`. Full suite passes (1338 passed, 21 skipped).

Found while investigating SCMDB blueprint import/export feasibility.

### DIFF
--- a/scripts/generate_enhancements_ini.py
+++ b/scripts/generate_enhancements_ini.py
@@ -88,6 +88,23 @@ except ImportError:  # pragma: no cover — only triggers if src/ is removed
     def win_long_path(path):  # type: ignore[misc]
         return str(path)
 
+# Same deferred-import pattern again. blueprint_meta.py's Blueprint Tracker
+# needs this exact same bp_craft_/bp_rewards_/bp_ prefix-strip when a raw
+# blueprint filename shows up verbatim in mission text; sharing it here
+# instead of a second hand-maintained copy means the two can't drift out of
+# sync on which prefixes are known.
+try:
+    _gen_root = Path(__file__).parent.parent
+    if str(_gen_root) not in sys.path:
+        sys.path.insert(0, str(_gen_root))
+    from src.utils.blueprint_meta import strip_raw_blueprint_filename_prefix
+except ImportError:  # pragma: no cover — only triggers if src/ is removed
+    def strip_raw_blueprint_filename_prefix(stem):  # type: ignore[misc]
+        for prefix in ("bp_craft_", "bp_rewards_", "bp_"):
+            if stem.lower().startswith(prefix):
+                return stem[len(prefix):], True
+        return stem, False
+
 
 # ── Paths ─────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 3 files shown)


Stealrull (commented):
## Self-review summary

Reviewed file-by-file (`src/utils/blueprint_meta.py`, `scripts/generate_enhancements_ini.py`, `tests/test_blueprint_meta.py`). Posting as a comment rather than an approval since GitHub doesn't allow approving your own PR.

**One Major finding, fixed during review:** `_raw_blueprint_filename_slug` was a near-verbatim copy of `generate_enhancements_ini.py`'s own `_name_from_blueprint_filename` transform — same prefix tuple, same strip+title-case logic, two hand-maintained copies that could drift apart. Extracted the shared piece into `strip_raw_blueprint_filename_prefix()` in `blueprint_meta.py`; `generate_enhancements_ini.py` now imports it using the same deferred-import + `ImportError`-fallback pattern already used for `tag_builder`/`win_paths`, so the standalone-CLI-without-`src/` case still works. No behaviour change for either caller, confirmed by both existing suites plus a full-suite run (1338 passed).

**Verified rather than assumed:**
- Checked out `blueprint_meta.py` from `upstream/release/2.3.0` (pre-fix) and confirmed the new regression test genuinely fails there (`KeyError: 'Norfield'`), then confirmed it passes with the fix — not a vacuous test.
- Checked the test's fixture claim against `tests/fixtures/kraken_global_latest.ini` directly — the real `UWC_Refueling_*` mission bodies do contain the exact raw-filename shape described (`bp_craft_nozzle_fuelgiver_grin_nozzlefast` etc., line 49620).

No Critical or remaining Major/Minor findings. CI passing. Ready to merge pending an actual approval.

Osiris-DevWorks (approved):
Verified the DRY extraction into blueprint_meta with the deferred-import fallback, the preserved fallback ordering (raw-filename slug slots after key-slug, before identity), and the non-vacuous regression test against the kraken fixture. One nit for the record: the helper matches prefixes case-insensitively where the old inline generator code was case-sensitive, so 'no behaviour change' is not literally true, though real blueprint filenames are lowercase so nothing observable shifts. Merging.

PR #336, opened 2026-08-01
Add blueprint import/export to the Blueprint Tracker

Stealrull:
## Stacked on #310

**This PR is stacked on #310** (which itself includes #268/#272). Please merge #310 first. The diff against `release/2.3.0` will show #310's commits until then; only the final commit belongs to this PR.

The dependency is real, not cosmetic:
- The Export/Import buttons sit in a 4-column grid alongside `_scan_other_channels_checkbox` (#272) and `_force_rescan_checkbox` (#310).
- The shared scan/import summary wording edits `_finish_blueprint_scan_queue`, which comes from #272.

Neither exists on `release/2.3.0` yet.

## What this adds

A community request (via Discord): let players move their owned-blueprint list between Smart Citizen and [scmdb.net](https://scmdb.net) instead of re-ticking hundreds of items by hand.

**Export** writes the Owned list as JSON or CSV, chosen from the save dialog's filter dropdown. The JSON shape matches what SCMDB itself emits (`version` / `exportedAt` / `missions` / `blueprints`) so it can be fed straight back in.

SCMDB's own `tag` and `url` fields are **omitted rather than faked** — Smart Citizen has no real value for them, and inventing one seemed worse than leaving it out. Happy to revisit if you'd rather they were present.

**Import** reads either format and is **strictly additive** — it never un-owns anything, so a partial or stale export can't silently wipe someone's list. It mirrors "Scan Logs for Owned Blueprints" exactly: no confirmation step, applied immediately, and the `[Owned]` tag is auto-woven into the strings table via `owned_items_changed` (hence `mark_owned_clean()`, not dirty — the re-weave already happened, same reasoning the scan path already uses).

Names in the file that Smart Citizen doesn't track are skipped and reported separately, so importing a full SCMDB dump shows what didn't carry over instead of silently dropping it. In real-world testing against a full SCMDB export, most non-matches are colour/skin armour variants that never appear in any mission's reward pool.

## Shared wording

Scan and import now share one message — "N blueprint(s) added to your Owned list. They'll show an `[Owned]` tag wherever they appear in a mission's blueprint rewards."

This replaces `enhancements.bp_scan_summary` and its "visible now" count, which described an implementation detail users had no way to act on. That key is removed from all languages.

## A note on `_relabel_details_button`

`QMessageBox.setDetailedText()` labels its toggle "Show Details...", which doesn't say what the details are. Renaming it turned out to need three separate workarounds — the dialog sizes its layout around the original short label, Qt rewrites the text on every click, and the button is not checkable so `isChecked()` can't drive the label. Each was confirmed by live testing rather than guessed at; the docstring records why, since none of it is obvious from the Qt docs.

## Testing

- 23 new tests in `tests/test_blueprint_export.py` covering export shape (JSON + CSV), import parsing (own format, genuine SCMDB files, malformed input), and name matching.
- Manually tested end-to-end in a portable build against a real ~1,500-entry SCMDB export.
- Full suite: 1369 passed, 21 skipped. Five `test_{chinese,italian,japanese}_activation` failures are **pre-existing on #310's branch** and unrelated (verified by running them on the unmodified base) — missing backfill keys and an `import_dialog.custom_value_prompt` placeholder drift.

## Translations

English is the source; the other 7 languages use AI translations pending human review, per the existing `{ht, at}` convention. **German is deliberately absent** because #313 hasn't merged yet — it'll need the usual backfill afterwards.


Stealrull (commented):
## Self-review summary

Reviewed file-by-file: `src/utils/blueprint_export.py`, `src/gui/blueprint_tracker_tab.py`, `src/gui/main_window.py`, `tests/test_blueprint_export.py`, and all 7 language files. Posting as a comment rather than an approval since GitHub doesn't allow approving your own PR. Note: the diff shown here includes #310's own commits since it isn't merged yet — this PR is genuinely one commit (now four, after review fixes) on top of it; see the PR description for the stacking details.

**Four findings, all fixed during review:**

1. **Major — confirmed live, this crashed the app:** `_parse_import_names_csv` only caught `OSError`, but a non-UTF-8 CSV (routine for a file saved from Excel on Windows, which defaults to cp1252/Latin-1) raised an uncaught `UnicodeDecodeError` that reached `sys.excepthook`. Reproduced the crash directly, fixed by also catching `UnicodeDecodeError`, then confirmed the same repro is caught cleanly post-fix. Regression test added.

2. **Minor — inaccurate docstring:** The test file claimed manual-only GUI testing "match[ed] every other QFileDialog-driven feature in this codebase." Checked, and it doesn't — `test_settings_profile.py`/`test_restore_backup.py` (covering #311, the most directly comparable prior feature) monkeypatch `QFileDialog` to unit-test their wiring directly. Corrected to point at that precedent instead of claiming it doesn't exist.

3. **Minor — asymmetric normalization, verified dormant but real:** `match_import_names` compared imported names (always normalized) against `blueprint_meta`'s own dict keys (never normalized), breaking `normalize_item_name`'s own documented "symmetric on both sides" invariant. Confirmed empirically this doesn't affect any of the 32 currently-known real blueprint names, but fixed it properly — normalizes a working copy of the known side for comparison, then returns each match's *original* key (not the normalized form), since the Owned set has to stay keyed exactly as `blueprint_meta`'s dict for later lookups to succeed. Added a test proving the old bare-intersection version would have missed this exact case.

4. **Minor — test gap:** `_parse_import_names_json`'s "`blueprints` key present but wrong type" branch (as opposed to "key absent entirely") had no direct test, despite being correct by inspection. Added one.

**Verified rather than assumed, throughout:**
- Hand-traced the save-dialog extension-detection logic across ~5 realistic cases.
- Checked every new i18n filter string's actual content against Qt's required format.
- Programmatically verified placeholder preservation (`{count}`, `{path}`, `{error}`, etc.) across all 6 translated languages against English — zero mismatches.
- Confirmed German's absence and the `bp_scan_summary` removal claim directly against what's pushed, not the PR description's word.
- Verified the `owned_items_changed` → `_recompute_owned` signal wiring the `mark_owned_clean()` reasoning depends on.

No remaining Critical/Major/Minor findings. Full suite: 1372 passed on the isolated branch (5 pre-existing i18n failures inherited from #310's unmerged base, unrelated to this work — verified by running the same tests against the unmodified base). Ready to merge pending an actual approval, and pending #310/#272 merging first per the stacking note.

Osiris-DevWorks (approved):
The import summary now counts only genuinely new names (matched minus owned), mirroring the log-scan flow, and skips the owned-set write and re-weave when nothing is new. All four self-review fixes verified earlier, the branch is updated against the current base, and the suite passes except the known German coverage gap, which #338 closes right behind this merge.

PR #338, opened 2026-08-01
i18n: backfill German for #311, #332, #303, #330, #336

Stealrull:
## Stacked on #313, #311, #332, #303, #330, #336

**This PR is stacked on six independent, unmerged PRs.** Please merge them first; the diff against `release/2.3.0` will show all six's commits until then — only the final commit belongs to this PR.

## What this fixes

German (#313) is the newest language and predates all five of the other PRs it's stacked on, so it's missing 72 keys total:

- The same 41 `config.*`/`enhancements.*`/`extract.*`/`settings_backup.*`/`status_bar.*` keys that #337 backfills for Chinese/Italian/Japanese — mostly #311's Export/Import Settings feature.
- 31 `blueprint_tracker.*` keys from #336's import/export feature, which explicitly excluded German for the same reason (it hadn't merged yet either — see #336's own PR description).
- `filters.ship_vehicle_names_only`/`_tooltip` from #330 — German-only gap, the other three AI languages already had it.

Also removes the orphaned `enhancements.bp_scan_summary` key that #336 replaced with the unified `owned_added_*` wording — same cleanup already done for chinese/italian/japanese as part of that PR's own review.

## Scope note

This branch is deliberately German-only. Running the full suite here shows two additional failures (`test_chinese_covers_full_english_key_universe`, `test_italian_covers_full_english_key_universe`) — those are **#337's scope**, a separate sibling PR, not something introduced or left unfixed by this one.

## Verification

All 9 `test_german_activation.py` tests pass. Full suite: 1503 passed, 21 skipped, only the 2 out-of-scope chinese/italian failures noted above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #337, opened 2026-08-01
i18n: backfill Chinese/Italian/Japanese for #311, #332, #303, #330

Stealrull:
## Stacked on #311, #332, #303, #330

**This PR is stacked on four independent, unmerged PRs.** Please merge them first; the diff against `release/2.3.0` will show all four's commits until then — only the final commit belongs to this PR.

## What this fixes

Chinese, Italian, and Japanese are each missing the same 39 translation keys, entirely because those four PRs were written before Chinese/Italian/Japanese existed (or the gap was simply out of scope for them at the time — same pattern as the earlier review feedback on #272/#310 in this cycle). Almost all of it (28 keys) is #311's Export/Import Settings feature (`settings_backup.*`, `config.backup_*`/`export_settings_*`/`import_settings_*`), plus smaller gaps from #332 (RS ore name annotations) and #303 (the p4k-locked message).

Japanese isn't caught by an automated key-coverage test the way chinese/italian are (`test_japanese_activation.py` doesn't have that specific check), so this gap stayed completely silent — the UI just quietly fell back to English for these 39 strings. Found it by checking directly rather than trusting the absence of a test failure.

## Verification

Checked comprehensively, not just the languages I expected to be affected: every language on this branch (chinese, french, italian, japanese, portuguese_br, spanish) now has **zero** missing keys and **zero** unexpected extras against English. Full suite passes (1454 passed, 21 skipped).

German needs this same backfill plus #336's blueprint import/export keys, but German (#313) hasn't merged yet — that'll be a separate PR stacked on top of it once it lands, so German doesn't fall further behind in the meantime.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #339, opened 2026-08-01
Fix Chinese/Italian mistranslating the literal [Owned] tag

Stealrull:
## What this fixes

Same bug as the German fix that came out of the blueprint import/export review cycle (#338): the `[Owned]` tag woven into mission text is a hardcoded literal —

```python
# src/utils/owned_items.py:29
_OWNED_TAG = " <EM4>[Owned]</EM4>"
```

— never translated per-language. Every player sees the literal English word `[Owned]` in their actual game files, regardless of what UI language Smart Citizen itself is set to.

Chinese translated the bracket text as `[已拥有]` in 4 places (`enhancements.blueprints_desc`, `enhancements.bp_scan_summary`, `blueprint_tracker.apply_owned_tag_tooltip`, `tutorial.blueprint_tracker.description`). Italian did the same as `[Posseduto]` in 3 of its 4 equivalent places — its `apply_owned_tag_tooltip` already had it right, inconsistently with the other three.

Both languages already had one correct reference each (`blueprint_tracker.owned_tags_refreshed`, from #297) that these mistranslations directly contradicted — a Chinese user reading `blueprints_desc` would expect a `[已拥有]` tag to appear, but `owned_tags_refreshed` (correctly) told them to expect `[Owned]`. Neither matched what's actually written into their game files except the correct one.

## Fix

Replaced the bracket text in all 7 occurrences (4 Chinese, 3 Italian) with the literal `[Owned]`, leaving the surrounding translated prose unchanged. Verified the exact occurrence count matched before and after the replace (no accidental over- or under-matching), and confirmed `owned_tags_refreshed` — the one string that was already correct in each language — is unaffected.

## Testing

Full suite: 1337 passed, 21 skipped, no failures. Both files re-validated as valid JSON post-edit.

Found while reviewing the German backfill PRs (#337/#338) for this same issue and checking whether it was isolated to German — it wasn't.


Stealrull (commented):
## Self-review summary

Reviewed both changed files. Posting as a comment rather than an approval since GitHub doesn't allow approving your own PR.

**No issues found.** Verified independently rather than trusting the PR body's own numbers:
- Diff checked line-by-line: every change touches only the bracket text (`已拥有`→`Owned`, `Posseduto`→`Owned`), surrounding translated prose is byte-identical before and after. Exactly 4 Chinese lines and 3 Italian lines changed, matching the claimed scope precisely.
- `owned_tags_refreshed` — the one string already correct in each language — confirmed absent from the diff, genuinely untouched.
- Checked out the actual PR branch and reran the full suite myself: 1337 passed, matching exactly.
- Placeholder integrity re-verified directly on `bp_scan_summary` (the one key carrying `{count}`/`{visible}`) — intact in both languages.

No Critical/Major/Minor findings. CI passing. Standalone PR, no stacking dependencies.

Osiris-DevWorks (approved):
Verified all 7 changes are bracket-text only with placeholders and surrounding prose intact, and that the in-code tag in owned_items.py is a hardcoded English literal, so these ui.json strings were descriptive drift, not matching logic. One heads-up: two of the edited strings are enhancements.bp_scan_summary, which #336 deletes, so this merges first to keep that PR clean. Merging.

Issue #31, opened 2026-05-14
Dynamic Tag Builder


Osiris-DevWorks:
Component Tag Builder 

In config tab, we'll allow users to configure the format of the component tag as they like in case they don't want to use the default. We will give options for:

Size Prefix
Class Text
Grade Text
Enclosing
Separators

In addition, we'll allow ordering of Size, Class, and Grade.

h0useRus:
Pattern could be
```
[%class(1)%size%grade] %name -> [M2A] FR-76
%class(3)-S%size-%grade %name -> CIV-S2-B FR-76
```

it could be simplified
```
[%class%size%grade] %name -> [M2A] FR-76
%class-S%size-%grade %name -> CIV-S2-B FR-76
```
where class gets from custom json mapping like, I would made class mapping (stored locally), because in my one letter version I use R(Race) for Completion
```json
// class_mapping.json
{
   "Military" : "M",
   "Civilian" : "C",
   "Industrial" : "I",
   "Competition": "R"
}
```

Osiris-DevWorks:
> Pattern could be
> 
> ```
> [%class(1)%size%grade] %name -> [M2A] FR-76
> %class(3)-S%size-%grade %name -> CIV-S2-B FR-76
> ```
> 
> it could be simplified
> 
> ```
> [%class%size%grade] %name -> [M2A] FR-76
> %class-S%size-%grade %name -> CIV-S2-B FR-76
> ```
> 
> where class gets from custom json mapping like, I would made class mapping (stored locally), because in my one letter version I use R(Race) for Completion
> 
> // class_mapping.json
> {
>    "Military" : "M",
>    "Civilian" : "C",
>    "Industrial" : "I",
>    "Competition": "R"
> }

We could extend the json values to be an array:

> // class_mapping.json
> {
>    "Military" : ("M", " MIL", "Military")
>    "Civilian" : ("C", "CIV", "Civilian"),
>    "Industrial" : ("I", "IND", "Industrial"),
>    "Competition": ("R", "CMP", "Competition)
> }

Though I'm thinking of also letting it be free-form text to give user full control.

h0useRus:
Also, please do the same for missiles, torpedoes and bombs, make sure you cover ground one ;)
```
[%type] %name

Marksman I Missile → [IR] Marksman I Missile
```
IR = Infrared Tracking
EM = Electromagnetic Tracking
CS = Cross-Section Tracking
B = Bomb

[...]

PR #340, opened 2026-08-01
Fix [BP] title tag ignoring chance-based blueprint rewards

Stealrull:
## Problem

The `[BP]`/`[BP?]` title-tag decision only checked whether a mission's variants *have* a `BlueprintRewards` pool at all — never whether that pool's own `chance` attribute is actually `1.0`. A mission whose blueprint reward is a coin flip (`chance="0.3"`, or any value below 1.0) still counted as `has_bp = True` with full confidence, so a title where every variant carries such a pool got tagged with the guaranteed `[BP]` even though you won't always get one.

**The mission DETAILS body already gets this right** — it reads the exact same per-variant `bp_chance` field and correctly shows "30% chance" instead of "Guaranteed" when `chance < 1.0`. So for the same mission, the title said guaranteed while the body said 30% chance. They disagreed about the same mission.

Reported against the Battaglia scan/mining contracts, which appear to be the first missions in the current data with a genuinely fractional `BlueprintRewards` chance — that's presumably why this gap was never hit before.

## Fix

Adds `_all_bp_guaranteed`: `_all_have_bp` is still required (every variant must have a pool at all), but now every one of those pools must also be `chance >= 1.0` before the title earns the unqualified `[BP]`. Anything short of that — partial coverage, surviving no-BP cargo descs (the existing #102/#31 demotion), or a sub-1.0 chance on an otherwise fully-covered title — demotes to the honest `[BP?]`, using the same demotion path already in place for the other two cases.

```python
_all_bp_guaranteed = _all_have_bp and all(
    v.bp_chance >= 1.0 for v in variants if v.has_bp
)
...
if _all_have_bp and _all_bp_guaranteed and not _surviving_no_bp_cargo:
    augmented_title += " <EM4>[BP]</EM4>"
elif _bp_partial or (
    _all_have_bp and (_surviving_no_bp_cargo or not _all_bp_guaranteed)
):
    augmented_title += " <EM4>[BP?]</EM4>"
```

## Testing

`tests/test_blueprint_pools.py::TestCargoBpTitleDemotion` tests the augment loop's tag branch via a replica (the real loop isn't separately callable — same pattern the existing `#102`/`#31` demotion tests already use). Updated the replica to match the fix and added 3 new tests:

- `test_full_coverage_with_partial_chance_demotes` — the Battaglia repro: full coverage, `chance=0.3` → `[BP?]`.
- `test_full_coverage_full_chance_stays_full_bp` — control case: full coverage, `chance=1.0` → still `[BP]` (confirms the fix doesn't just always demote).
- `test_mixed_chance_across_variants_demotes` — one guaranteed variant, one 50/50 variant sharing a title → `[BP?]`, the guaranteed one doesn't get to win.

Confirmed the new test genuinely reproduces the bug by running it against the pre-fix logic first (produced `[BP]`, the wrong answer) before applying the fix.

Full suite: 1340 passed, 21 skipped, no failures.

---

Related to #341, but does not close it. Investigation against the real LIVE DataForge cache (2026-08-01): every Battaglia `BlueprintRewards` is `chance="1"` with a pool on every variant, so the non-guaranteed payouts players see there come from pool state (a pool only awards blueprints you don't own yet), not from anything static mission text can express. The chance-attribute mislabel this PR fixes is real but lives in Rayari's `RAIN_collectresources` missions (`chance="0.25"`): a full generator run on this branch flips exactly those two titles from [BP] to [BP?] and changes nothing else, Battaglia included.


Stealrull (commented):
## Self-review summary

Reviewed both files. Posting as a comment rather than an approval since GitHub doesn't allow approving your own PR.

**One Minor finding, addressed during review:** the `elif` branch's `not _all_bp_guaranteed` term is logically redundant in production today — proved algebraically that `_all_have_bp = True` implies `_has_dominant_no_bp_bucket = False` implies `_bp_partial = True` unconditionally, so `_bp_partial` alone already catches this branch whenever the `if` fails on the new guard. It's only load-bearing in the test replica, which computes `_bp_partial` differently (a simplified `not _all_have_bp`, the opposite relationship). Not a bug — both paths reach the same correct answer — but added a comment explaining why the term is kept explicit anyway (defensive against a future rework of `_bp_partial`'s dominant-bucket heuristic) rather than removing it as dead code.

**Verified rather than assumed:**
- Traced the `if`/`elif` short-circuit structure by hand to confirm exactly how the bug manifested (the `if` branch matched first regardless of chance, before the fix).
- Proved the elif-redundancy claim against the actual `_has_dominant_no_bp_bucket` code, not just reasoned about it abstractly.
- Confirmed the regression test genuinely reproduces the bug against the pre-fix logic before the fix was applied.
- Full suite: 1340 passed on the isolated branch.

No Critical/Major findings. CI passing.

Osiris-DevWorks (approved):
Ran the investigation the author couldn't finish: the fix's logic is correct, but the live repro is Rayari, not Battaglia. Verified against the real LIVE DataForge cache: Battaglia's BlueprintRewards are all chance=1 (their in-game non-payouts are pool-state, not data), while Rayari's RAIN_collectresources missions carry chance=0.25 and were wrongly tagged [BP]. A full generator run on this branch flips exactly those two titles to [BP?] and changes nothing else. Comments and docstrings now attribute the repro correctly, and the PR no longer auto-closes #341. Merging.

PR #355, opened 2026-08-06
Fix Blueprint Tracker ignoring Tag Builder enclosing styles (#352)

### LINKED ISSUE #352: v2.3.0 Blueprint Tracker ignores Ship Weapons Tag Builder formatting on portable version
Tested again with a clean extraction of the v2.3.0 portable ZIP.

I regenerated and applied enhancements from scratch.
The status bar shows 89,871 entries and 75 overrides.

Component tags are displayed correctly in Blueprint Tracker, for example:
5CA 'Akhura' <CIV-S3-C-Shield>
7CA 'Nargun' <CIV-S3-A-Shield>

Ship weapons remain untagged, for example:
10-Series Greatsword Cannon
11-Series Broadsword Cannon
9-Series Longsword Cannon

The Ship Weapons Tag Builder preview is correct, but its formatting is not
used by Blueprint Tracker. This also occurs after a completely clean copy.

Stealrull:
Fixes #352.

## Summary
- Only the default Square `[]` enclosing style ever got stripped back off a tagged item name for Blueprint Tracker matching. Round `()`, Curly `{}`, Angle `<>`, and None (space only) all left Ship Weapons and the manual QuadraCell/QuadraCell MT/FR-66/FR-76 entries untagged with no Class/Size/Grade facets, as reported in #352.
- Round/Curly/Angle: thread the actually-configured enclosing through `normalize_item_name`/`parse_component_tag` instead of hardcoding Square, sourced from the live Tag Builder config via a new `enclosings_from_tag_configs` helper. Threaded through the Blueprint Tracker matching path, the main table's Owned-star column, and SCMDB import matching.
- None (space only) has no delimiter char, so bracket stripping structurally can't apply to it. Recovered instead by diffing a tagged value against the item's known stock (pre-Tag-Builder) value -- already available as `default_values` from the cached `base.ini` -- which is authoritative rather than a guess and works regardless of the tag's shape or spacing. Falls back to a conservative heuristic (an embedded Tag-Builder separator char plus a strict size/grade-shaped token) for contexts with no stock value to diff against, e.g. mission bullet text and the hardcoded manual items.
- An earlier attempt at the None-style fix inserted an extra marker space into generated game files at Tag Builder render time; reverted after it collided with real messy whitespace already present in some raw item names (locked in as a regression test). The final approach makes no changes to generated data at all.

## Test plan
- [x] Full test suite: **1618 passed, 21 skipped, 0 failed** (`pytest -q`) - includes the 3 tests added for the gate fix below
- [x] New/updated coverage: enclosing-aware stripping for Round/Curly/Angle, the `enclosings_from_tag_configs` category/fallback rules, the stock-diff recovery path and its no-match/empty-stock edge cases, the heuristic fallback and its two collision-safety regression guards (`"F55  LMG   Magazine"` messy whitespace, `"10-Series Greatsword Cannon"` real hyphenated name), and an end-to-end `build_blueprint_metadata`/Owned-star-column check for each enclosing style including None.
- [x] Manual portable build test: reporter (maintainer) confirmed all 5 enclosing styles now correctly tag Ship Weapons/QuadraCell in both the app and in-game, including the previously-unfixable None (space only) case.

---

**Update (27 Aug):** your Major -- `parse_component_tag` running the None-style heuristic ungated -- is fixed in `4e5d04c`, gated identically to `normalize_item_name`, with the exact tests you specified. Your 7 Minor items are untouched, as deferred. See the comment thread for that and for the #355/#358/#375 merge-order note.


Osiris-DevWorks (commented):
Thanks for the fix. The overall design is right: `owned_items` stays settings-free, callers resolve the live config through `enclosings_from_tag_configs` and pass plain tuples down, and the Square-only default is guarded by explicit regression tests. One question to settle and two small cleanups before merge.

**Question: the None-style heuristic is gated in `normalize_item_name` but ungated in `parse_component_tag`.** `normalize_item_name` only runs the space-only heuristic when the None style is actually configured, exactly because real names like "F-4 Blaster" can look tag-shaped. `parse_component_tag` runs the same `find_none_style_tag_word` fallback unconditionally, and `test_parse_component_tag_none_enclosing_via_heuristic_without_stock` blesses that. So a default Square-only user whose component value has no stock match (user-renamed values, keys missing from `default_values`) can now get fabricated Class/Size/Grade facets where the old code returned nothing, which contradicts the "default behavior preserved exactly" claim. Exposure is small since stock diffing covers almost everything, so this is a question, not a blocker: either gate the fallback on `NONE_STYLE_ENCLOSING` being in the enclosings tuple for symmetry (and update that test), or add a comment saying why it is deliberately looser here.

**Fix: the comment in `scan_entity_dir` is wrong.** It says `join_tag` "sniffs the tag string itself for its enclosing style"; `join_tag` just joins with a space. Looks like a leftover from the reverted marker-space attempt. Just correct the comment.

**Note:** there are now two different `_build_tag_patterns` functions with the same name (`owned_items.py` strips, `blueprint_meta.py` captures). Two occurrences is within our DRY tolerance, but the shared name invites confusion; a rename or a cross-reference comment in each would do.

Osiris-DevWorks (changes requested):
## Review — #352 enclosing styles

Strong PR. The architecture is right: `owned_items` and `blueprint_meta` stay Qt-free and settings-free, taking a plain `enclosings` tuple while the GUI resolves live config. The stock-diff approach is genuinely better than guessing the tag boundary, and the reverted marker-space attempt is documented so nobody retries it. Threading is complete — I enumerated every caller of the six changed functions across `src/` and `scripts/`; none were missed.

**Verification:** full suite **1616 passed, 21 skipped, 0 failed** (Python 3.14.7 / PyQt6 6.11.0). Findings below were confirmed by executing the branch, not by reading.

**Acceptance against #352** — the reporter's own examples under Angle:

```
'5CA 'Akhura' <CIV-S3-C-Shield>'         -> name "5CA 'Akhura'"              facets (CIV, S3, C)
'<B-S2-A> 10-Series Greatsword Cannon'   -> name '10-Series Greatsword Cannon'  facets (-, S2, B)
```

Both resolve, including the exact ship-weapon case from the issue. Class reads as `None` on `<B-S2-A>` because a single-letter class is ambiguous with a grade — the pre-existing Short-style limitation, not a regression.

---

### Major — `parse_component_tag` runs the None-style heuristic ungated

`blueprint_meta.py:411-414`. Its sibling `normalize_item_name` gates on `NONE_STYLE_ENCLOSING in resolved`; this one doesn't. Same input, default Square-only config, None configured nowhere:

```
value                    normalize_item_name        parse_component_tag
250-E Laser Pointer      '250-E Laser Pointer'  OK   (None, 'S250', 'E')  <- fabricated
C-788 Cannon             'C-788 Cannon'         OK   (None, 'S788', 'C')  <- fabricated
A-18 Shredder            'A-18 Shredder'        OK   (None, 'S18',  'A')  <- fabricated
```

A populated `stock` short-circuits it, so English installs are mostly covered. It is **systematic on non-English installs**, where `default_values` is English by design — the diff never matches and the ungated heuristic runs on every translated name:

```
value='C-788 Canon'  stock='C-788 Cannon'  ->  (None, 'S788', 'C')
```

This is the same defect class your own review rated Major and fixed in 3d0f06c; it just didn't reach the second call site. Facets only, so it cannot produce a false `[Owned]`.

Fix mirrors the sibling (hoisting `resolved` so it's in scope):

```python
if inner is None and NONE_STYLE_ENCLOSING in resolved:
    found = find_none_style_tag_word(value or "")
```

**Related test gap — this is why the suite stayed green.** `test_none_style_heuristic_is_off_unless_configured` has no `parse_component_tag` counterpart, and `test_parse_component_tag_none_enclosing_via_heuristic_without_stock` asserts the ungated behaviour while its docstring claims parity with `normalize_item_name`'s fallback. That test needs its config argument added and the docstring corrected, plus a new gate test:

```python
def test_parse_component_tag_heuristic_is_off_unless_none_configured():
    assert parse_component_tag("250-E Laser Pointer") == (None, None, None)
    assert parse_component_tag("C-788 Cannon") == (None, None, None)
    none_cfg = (("[", "]"), NONE_STYLE_ENCLOSING)
    assert parse_component_tag("MIL-S3-B Balandin", none_cfg) == ("MIL", "S3", "B")
```

---

### Minor

1. **Stale comment.** `generate_enhancements_ini.py:5792` says *"join_tag sniffs the tag string itself for its enclosing style"*. It doesn't — it takes `placement` and joins with a space. Leftover from the reverted approach; the `tagger` note below it is still valid and worth keeping.

2. **`_place_commodity_tag` fallback inverted.** Old: prepend only on `"prepend"`, everything else appends. New: append only on `"append"`, everything else prepends. `TagConfig.from_dict` normalizes to `prepend`/`append`/**`replace`**, so `"replace"` flips this. The commodities editor is wired to `PLACEMENTS` (prepend/append only), so it needs a hand-edited config to reach — cheap to guard anyway.

3. **DRY — `enclosings_from_tag_configs(AppSettings.get_all_tag_configs())` appears 6x** across `main_window.py` and `blueprint_tracker_tab.py`. Root CLAUDE.md extracts at three. An `AppSettings.get_tag_enclosings()` helper keeps the category list (`components`/`missiles`/`ship_weapons`) in one editable place — right now changing it is invisible from six sites.

4. **`_DEFAULT_ENCLOSINGS` defined identically** in `owned_items.py:41` and `blueprint_meta.py:288`. The magic-literal rule extracts at two; `blueprint_meta` already imports five names from `owned_items`.

5. **`_build_tag_patterns` near-duplicated** in both modules, differing only in capture group and `*` vs `+`. Judgment call given the two-occurrence rule, but a future enclosing style means editing both.

6. **Private cross-module import** — `from src.utils.tag_builder import _ENCLOSING_BY_KEY` (`owned_items.py:799`). `ENCLOSINGS` is public at `tag_builder.py:137`.

7. **Unannotated params** — five bare `enclosings=None` in `blueprint_export.py`, one in `string_table_model.py`, against `"Sequence[tuple[str, str]] | None"` everywhere else.

---

### Noted, no action needed

**Paint-path double-normalize.** `_is_bp_item` calls `_owned_name`, then the caller calls it again — 4x per visible Owned cell per repaint (`string_table_model.py:385-387`, `415-416`, `426-427`). Same argument you applied to `extract_bp_item_names` in this PR. The table is virtualized so it's tens of rows, and I didn't profile it — only raising it for consistency.

**The Round trade-off.** I initially read the `(Modified)` collision as a bug before finding `test_configured_non_square_enclosing_can_collide_with_real_parenthetical_name`. It's deliberate and test-locked, and I checked that the obvious fix doesn't work: a class-only tag renders as `(Military)`, which a separator+size/grade shape check would refuse to strip, breaking matching outright. Blanket stripping is the safer default.

One data point for a future ticket, not this PR: in `kraken_global_latest.ini`, 300 names end in a parenthetical and **48 have a bare-name twin that is a separate real item** — `Artimex Arms` / `Artimex Arms (Modified)` are adjacent bullets in the same POTENTIAL BLUEPRINTS list. If it's ever worth closing, the principled route is building the strip pattern from the actual `TagConfig` (elements and separator are known) rather than reducing it to a bare bracket pair.

---

### Merge order

As you flagged, this and #358 both add a different keyword to `extract_bp_item_names`, `apply_owned_to_value`, `build_blueprint_metadata` and `filter_entry_indices` — the second to merge conflicts in four files. The parameters are orthogonal; keep both. Worth deciding the order before either lands.

---

Requesting changes on the Major item and its test only. Everything under Minor is polish and I'd happily see it deferred.

Stealrull (commented):

Reviewed the full diff against Smart Citizen's layer ordering and standards, plus the two rounds of review already on this PR.

**Blocking item — resolved.** Osiris's Major (`parse_component_tag` running the None-style heuristic ungated, fabricating facets from ordinary names like `250-E Laser Pointer`) is fixed in the current diff, gated identically to `normalize_item_name`. Confirmed the exact tests he asked for are present and passing: `test_parse_component_tag_heuristic_is_off_unless_none_configured` (all 3 of his fabrication examples), `test_parse_component_tag_gate_holds_on_a_non_english_install`, and the corrected docstring/config on `test_parse_component_tag_none_enclosing_via_heuristic_without_stock`.

**Minor items** — Osiris's 7 Minor findings (stale `join_tag` comment, `_place_commodity_tag` fallback inversion, DRY opportunities, private cross-module import, unannotated params) remain as filed. He explicitly called these polish and said he'd be happy to see them deferred, so treating them as non-blocking here too.

**Fresh pass:** checked for direct `QSettings()` use, hardcoded column indices vs `COL_*`, and unsafe `self.entries[idx]` access introduced by this diff -- none found. `join_tag`'s extraction is a solid DRY win across the 4 `generate_enhancements_ini.py` call sites. Test coverage is thorough (enclosing round-trips per style, the stock-diff path, the gate itself, the manual-item facet resolution under non-Square styles).

Nothing new to add. No blocking issues from this pass.

Osiris-DevWorks (dismissed):
## Re-review — blocker resolved, approving

Lifting my earlier Request Changes. `Gate the space-only heuristic in parse_component_tag too` (2026-08-26) fixes it properly, and I verified it on the branch rather than reading the diff:

```
default config (None style NOT configured anywhere):
  250-E Laser Pointer  -> (None, None, None)      was (None, 'S250', 'E')
  AVS-E Helmet         -> (None, None, None)      was ('AVS', None,  'E')
  C-788 Cannon         -> (None, None, None)      was (None, 'S788', 'C')
  A-18 Shredder        -> (None, None, None)      was (None, 'S18',  'A')

non-English install (English stock vs translated value) -- the systematic gap:
  parse_component_tag('C-788 Canon', SQUARE, 'C-788 Cannon')  -> (None, None, None)

still parses a real None-style tag when the style IS configured:
  parse_component_tag('MIL-S3-B Balandin', (SQUARE, NONE_STYLE_ENCLOSING))  -> ('MIL', 'S3', 'B')

suite: 1618 passed, 21 skipped, 0 failed
```

Both halves hold: the fabrication is gone in every case I raised, including the non-English path, and the gate doesn't make recovery inert for the users who actually chose None. The three tests I asked for are present.

The 7 Minor items stand as filed and remain non-blocking — the stale `join_tag` comment is the one I'd still bother with, since it describes behaviour the function doesn't have and will mislead whoever touches ship-weapon tagging next.

### Merge order

I test-merged all seven open 2.3.1 PRs to check your analysis. It is exactly right, including the hunk counts:

```
each PR onto release/2.3.1 individually:   all 7 clean
#371 + #365 + #376 + #377 together:        clean at every step
#355 after #358:  12 hunks / 3 files -> owned_items.py 6, blueprint_meta.py 3, main_window.py 3
```

6/3/3 matched on the nose, and "keep both" is the right resolution in every case.

Two adjustments to the plan:

**#376 should come out of the free-merge group** — I filed Request Changes on it after your message. Filtering the Owned list means a stale selection under a filter lets Remove un-own items the user can't see (`selects 3 -> filters -> visible=['Colossus'] -> Remove unowns all 3`). One line in `_selected_names`.

**Your `known_item_names()` follow-up is more than a consistency fix.** On a non-Square install the catalogue would key under Square-only rules while owned names key under the configured style, so more names read as unmatched — which feeds the drop path I flagged on #375, where an unmatched name resolving onto an already-owned entry is deleted outright. The forwarding fix and "keep both instead of dropping" reinforce each other; worth doing both.

I'd also still put **#377 last** rather than free — every section describes still-open PRs, so landing it early risks a `build-installer` preview handing testers a checklist for work that isn't in the build.

PR #356, opened 2026-08-06
Fix Blueprint Tracker "stacking" fake items from commodity data (#354)

### LINKED ISSUE #354: Blueprint tracker 'stacks' components
Image

Reported by swill008 on discord, blueprint tracker looking as if it stacks some blueprints in the owned section.

Stealrull:
Fixes #354.

## Summary
- Reported via Discord: the Blueprint Tracker's Available and Owned lists were "stacking" fake entries like `"Power Plants: 10 items"` and `"Quantum Drives: 3 items (S1-S3)"` alongside real items.
- Root cause: `has_bp_section()` in the Blueprint Tracker's matching layer (`blueprint_meta.py`) ran on **every entry's value unconditionally**, regardless of category. Commodity items (e.g. raw ores) get a separate, independently-renameable "Blueprint Data" section listing what they're used to craft. If that header text collides with the mission "Blueprints" header (e.g. both left at/set to the same text), the commodity's crafting-material-usage summary gets scanned as if it were a real mission's blueprint reward list -- fabricating fake, unownable "items" indistinguishable from real ones in both Available and Owned.
- Fix: gated blueprint-bullet scanning to `category == "Missions"` entries only, since a real POTENTIAL BLUEPRINTS section never legitimately exists anywhere else. Applied consistently to both the Blueprint Tracker's matching (`blueprint_meta.py`) and the String Editor's "BP Descriptions" filter (`entry_filter.py`), which had the identical unscoped check.
- Every existing test with blueprint-bearing content already used `category="Missions"` in its fixtures, confirming this gate matches the codebase's own intent -- it just wasn't enforced in code yet.

## Test plan
- [x] Full test suite: 1577 passed, 0 failed (`pytest -q`)
- [x] New coverage: a commodity entry whose value contains a colliding "POTENTIAL BLUEPRINTS"-shaped header is confirmed excluded from both `build_blueprint_metadata()`'s output and the String Editor's "BP Descriptions" filter.
- [x] Manual portable build test: reproduced the exact collision (set "Blueprint data" to the same text as "Blueprints", left everything else at default) and confirmed the tracker no longer stacks fake entries, while real mission-derived blueprints continue to populate normally.

### DIFF
--- a/src/gui/main_window.py
+++ b/src/gui/main_window.py
@@ -50,7 +50,9 @@ from src.gui.workers import (
     StartupSyncWorker,
 )
 from src.merger.ini_merger import merge_sources_by_hierarchy
-from src.models.string_model import StringEntry, is_favoritable_ship
+from src.models.string_model import (
+    CATEGORY_MISSIONS, StringEntry, is_favoritable_ship,
+)
 from src.parser.ini_parser import load_source_files, load_sources_from_settings, parse_ini_file
 from src.gui.update_dialog import UpdateDialog
 from src.utils.app_updater import AppUpdateCheckWorker, AppUpdateDownloadWorker
@@ -5008,7 +5010,7 @@ class MainWindow(QMainWindow):
         entry_categories = set(e.category for e in self.entries)
 
         # Always include standard categories, even if no entries exist for them yet
-        standard_categories = {"Ships", "Ship Items", "Missions", "Commodities", "Other"}
+        standard_categories = {"Ships", "Ship Items", CATEGORY_MISSIONS, "Commodities", "Other"}
         categories = sorted(standard_categories | entry_categories)
 
         self.category_combo.blockSignals(True)
--- a/src/models/string_model.py
+++ b/src/models/string_model.py
@@ -86,6 +86,23 @@ _MISSION_PREFIXES = (
     "wantedlevel", "wstr_", "xenothreat_",
 )
 
+# The category name for mission entries. Named because #354 turned it from a
+# display string into a load-bearing gate: blueprint_meta and entry_filter both
+# refuse to scan an entry for a POTENTIAL BLUEPRINTS section unless it carries
+# this exact category, so a typo in either would silently empty the Blueprint
+# Tracker rather than fail loudly. Every place that decides "is this a mission
+# entry" reads this constant: the classifier below, ini_parser's source-based
+# override, and the two scan gates.
+#
+# Deliberately not localized. This is an internal classification token, not UI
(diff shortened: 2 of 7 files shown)


Stealrull (commented):
## Review

**Overview:** Small, focused fix. `has_bp_section()` (detects a mission's "POTENTIAL BLUEPRINTS" reward section) was checked against every loaded entry's value with no category scoping. A commodity item's independently-configurable "Blueprint Data" header (crafting-material-usage summary, e.g. `"- Power Plants: 10 items"`) could collide with the mission "blueprints" header text, causing that commodity text to be scanned as a real mission reward list -- fabricating fake, unownable items that appeared identically in Available and Owned ("stacking"). Fix: gate the scan to `category == "Missions"` entries only, applied consistently in `blueprint_meta.py` and `entry_filter.py`.

**Code quality:** Consistent with codebase conventions (no em dashes, comment style, docstring cross-references). Correctly scoped -- the fix lives at the two call sites with category context rather than pushing a `category` param into `owned_items.py`'s intentionally generic, Qt-free `has_bp_section`/`_bp_section_span`. Good architectural restraint.

**Risk worth flagging (non-blocking):** The fix's correctness now depends on category classification being complete. Before this change, `has_bp_section()` didn't care what category an entry was. After, only entries already classified `"Missions"` (via `_MISSION_PREFIXES` key-prefix matching) are eligible at all. If any real mission's key shape isn't covered by `_MISSION_PREFIXES` and falls into the `"Other"` bucket, that mission's blueprint section would now be silently excluded from the tracker entirely -- a new, silent failure mode that didn't exist before (previously category was irrelevant to blueprint scanning). Not a reason to hold this PR -- the collision bug it closes is worse -- but worth watching for reports of "missions missing from the tracker" post-release, since that would point here rather than back to #354.

**Test coverage:** Solid -- new tests confirm a colliding commodity entry is excluded from both `build_blueprint_metadata()` and the "BP Descriptions" filter. Existing fixture updates (adding `category="Missions"`) are a necessary, correctly-scoped side effect, not incidental.

**Performance/security:** No concerns -- the added check is a cheap string comparison already available in-loop.

Osiris-DevWorks (approved):
Thanks for the fix. I verified the load-bearing claim independently: every category-assignment route for mission descriptions lands on "Missions", the token is never localized, and the new `CATEGORY_MISSIONS` constant covers every gate site, with the `ENHANCEMENT_LABELS["missions"]` equality pinned by a test. The colliding-header commodity case is covered on both the metadata and filter sides. The three self-review findings are all resolved in the branch. No findings from this pass.

PR #358, opened 2026-08-07
Fix Blueprint Tracker ignoring a renamed "blueprints" mission header (#353)

### LINKED ISSUE #353: blueprint tracker doesn't populate if "POTENTIAL BLUEPRINTS" is renamed.
User reported via discord, if a user renames "POTENTIAL BLUEPRINTS" then the blueprint tracker will not populate as it looks for that header specifically, need to take into account users renaming that title in the mission labels section.

Stealrull:
Fixes #353.

## Summary
- Reported: renaming the mission "Blueprints" section header (Config > Mission Labels) silently breaks the Blueprint Tracker -- it stops populating entirely, not just mis-tagging.
- Root cause: the generator writes the user's actual configured header text into mission bodies instead of the hardcoded default, but `has_bp_section`/`_bp_section_span` (the "is this a blueprint-bearing mission body" gate) only ever recognized the literal `"POTENTIAL BLUEPRINTS"` default.
- Fix: thread the real configured header (`AppSettings.get_mission_headers()["blueprints"]`) through the matching layer -- `owned_items.py`'s `has_bp_section`/`_bp_section_span`/`extract_bp_item_names`/`apply_owned_to_value`, `blueprint_meta.py`'s `build_blueprint_metadata`, and `entry_filter.py`'s "BP Descriptions" filter -- mirroring the pattern already used for #352's configurable enclosing style.
- **Follow-up fix found during manual testing:** the header match must be anchored to the actual `<EM3>`/`<EM4>` wrapper the generator always uses (`f"<{em}>{header}</{em}>"`), not a bare substring search anywhere in the mission body. `"POTENTIAL BLUEPRINTS"` is distinctive enough to never collide with real mission flavor text by accident, but a user-renamed header can easily be an ordinary word -- reported case: renamed to `"stuff"`, which already appeared as plain prose earlier in the same mission body (a stray NPC/contact name list: "Ed", "H", "Wallace", ...). The old unanchored match found that earlier, unwrapped occurrence and started the blueprint-bullet scan from there, sweeping those unrelated names into the item set as if they were real blueprints. Anchoring to the properly-tagged header closes this.

## Test plan
- [x] Full test suite: **1629 passed, 21 skipped, 0 failed** (`pytest -q`) - re-run after the #356 merge landed on this branch
- [x] New coverage: renamed-header recognition (with and without the `bp_header` param, case-insensitivity, default-still-works-when-not-renamed), and a direct regression test reproducing the reported collision (a renamed header word appearing unwrapped in prose earlier in the same mission body, followed by the real EM-wrapped header).
- [x] Manual portable build test, two rounds: first round reproduced the "stuff" collision bug live; second round (after the anchoring fix) confirmed the fake entries are gone and the Blueprint Tracker populates correctly with a fully renamed header.

### DIFF
--- a/src/gui/main_window.py
+++ b/src/gui/main_window.py
@@ -1728,7 +1728,7 @@ class MainWindow(QMainWindow):
             _owned = AppSettings.get_owned_items()
             if _owned:
                 from src.utils.owned_items import apply_owned_to_value
-                _bp_header = AppSettings.get_mission_headers().get("blueprints")
+                _bp_header = self._bp_header()
                 for _k, _v in list(merged_dict.items()):
                     _nv = apply_owned_to_value(_v, _owned, bp_header=_bp_header)
                     if _nv != _v:
@@ -4996,7 +4996,7 @@ class MainWindow(QMainWindow):
             bp_titles_only=self.bp_titles_check.isChecked(),
             bp_descs_only=self.bp_descs_check.isChecked(),
             ship_vehicle_names_only=self.ship_vehicle_names_only_check.isChecked(),
-            bp_header=AppSettings.get_mission_headers().get("blueprints"),
+            bp_header=self._bp_header(),
         )
 
     @timed
@@ -5278,6 +5278,24 @@ class MainWindow(QMainWindow):
         self.table.setCurrentIndex(custom_index)
         self.table.edit(custom_index)
 
+    def _bp_header(self) -> "str | None":
+        """The user's configured "blueprints" mission header, or None when it
+        has never been renamed (#353).
+
+        Every matcher that has to recognise a blueprint section needs this, and
+        all of them must agree: the metadata scan, the owned re-weave, the
+        String Editor's BP filter, and the Apply-to-Game write path. Reading it
+        through one accessor is what keeps a fifth caller from being added with
+        a different key by mistake.
+
+        Read fresh on every call rather than cached. The header is editable in
+        the Enhancements tab mid-session, so a cached copy would need
+        invalidating from that edit, and a stale one means the tracker quietly
+        stops matching the text the generator is now writing. The read is a
+        settings lookup on paths that already do far more work than that.
+        """
+        return AppSettings.get_mission_headers().get("blueprints")
+
     def _rebuild_blueprint_metadata(self):
         """#157 follow-up: scan loaded strings once to build the blueprint-item
         metadata (eligible names + per-item mission/type/class/size/grade) the
@@ -5285,7 +5303,7 @@ class MainWindow(QMainWindow):
         it runs on load — not on every owned-toggle (that path re-partitions
         the cached result)."""
         from src.utils.blueprint_meta import build_blueprint_metadata
-        bp_header = AppSettings.get_mission_headers().get("blueprints")
+        bp_header = self._bp_header()
         self._blueprint_meta = build_blueprint_metadata(self.entries, bp_header=bp_header)
         self._bp_item_names = set(self._blueprint_meta)
 
@@ -5296,7 +5314,7 @@ class MainWindow(QMainWindow):
         the tag. The eligible-name set + filter metadata are built separately by
         `_rebuild_blueprint_metadata` (cached, not rescanned here)."""
         from src.utils.owned_items import apply_owned_to_value
-        bp_header = AppSettings.get_mission_headers().get("blueprints")
+        bp_header = self._bp_header()
         owned = AppSettings.get_owned_items()
         for e in self.entries:
             new_val = apply_owned_to_value(e.original_value, owned, bp_header=bp_header)
--- a/src/gui/main_window.py
+++ b/src/gui/main_window.py
@@ -1728,8 +1728,9 @@ class MainWindow(QMainWindow):
             _owned = AppSettings.get_owned_items()
             if _owned:
                 from src.utils.owned_items import apply_owned_to_value
+                _bp_header = AppSettings.get_mission_headers().get("blueprints")
                 for _k, _v in list(merged_dict.items()):
-                    _nv = apply_owned_to_value(_v, _owned)
+                    _nv = apply_owned_to_value(_v, _owned, bp_header=_bp_header)
                     if _nv != _v:
                         merged_dict[_k] = _nv
 
@@ -4995,6 +4996,7 @@ class MainWindow(QMainWindow):
             bp_titles_only=self.bp_titles_check.isChecked(),
             bp_descs_only=self.bp_descs_check.isChecked(),
             ship_vehicle_names_only=self.ship_vehicle_names_only_check.isChecked(),
+            bp_header=AppSettings.get_mission_headers().get("blueprints"),
         )
 
     @timed
@@ -5283,7 +5285,8 @@ class MainWindow(QMainWindow):
         it runs on load — not on every owned-toggle (that path re-partitions
         the cached result)."""
         from src.utils.blueprint_meta import build_blueprint_metadata
-        self._blueprint_meta = build_blueprint_metadata(self.entries)
+        bp_header = AppSettings.get_mission_headers().get("blueprints")
+        self._blueprint_meta = build_blueprint_metadata(self.entries, bp_header=bp_header)
         self._bp_item_names = set(self._blueprint_meta)
 
     def _recompute_owned(self):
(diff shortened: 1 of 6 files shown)


Stealrull (commented):
## Review

**Overview:** Two-part fix, the second part found via manual testing of the first. (1) Core fix: the "blueprints" mission section header is user-configurable (Config > Mission Labels), and the generator writes the user's actual renamed text into mission bodies -- but `has_bp_section`/`_bp_section_span` only ever recognized the hardcoded "POTENTIAL BLUEPRINTS" default. Threaded the real configured header through the full matching chain (`owned_items.py`, `blueprint_meta.py`, `entry_filter.py`, sourced in `main_window.py`). (2) Follow-up (found live): the header match wasn't anchored to the `<EM3>`/`<EM4>` wrapper the generator always uses -- it was a bare substring search anywhere in the mission body. A distinctive phrase like "POTENTIAL BLUEPRINTS" never collided by accident, but a short user-chosen replacement (reported: "stuff") can coincidentally appear as ordinary prose earlier in the same mission, causing the scan to start from that false match and sweep unrelated text (a stray NPC name list) in as fake blueprint items. Fixed by requiring the match to occur inside a properly matched `<EM([34])>...</EM\1>` pair.

**Code quality:** Consistent with codebase conventions throughout. Good instinct catching and fixing the anchoring gap rather than shipping the "technically adds the parameter" version -- the reported collision is a direct consequence of the design, not an edge case. The docstring is thorough about why the anchoring matters, including the concrete real-world trigger. Removing the now-dead `_BP_HEADER_RE` constant instead of leaving it unused is good hygiene.

**Worth noting (non-blocking):**
- The anchoring change is technically broader than the stated bug -- it also retroactively tightens matching for the two built-in default headers, not just the new renamed-header case. Very likely safe (existing test fixtures, including ones derived from real CIG data, confirm both headers are always EM-wrapped in practice), but worth being aware of if anything unexpected surfaces for users who haven't renamed anything.
- Residual, inherent risk: a very short/generic custom header could still, in principle, coincidentally match a properly EM-wrapped occurrence of the same text elsewhere in a mission body. Much narrower window than the bug just fixed, not practically fixable without deeper structural parsing -- reasonable to accept, same category of trade-off already documented for #352's enclosing-style collisions.

**Test coverage:** Strong. Covers renamed-header recognition with/without `bp_header`, case-insensitivity, default-still-works-when-not-renamed, and a direct regression test reproducing the exact reported collision -- which would have caught this before it shipped. The one pre-existing test using an unrealistic unwrapped-header fixture was correctly updated to realistic EM-wrapped text.

**Performance/security:** No concerns. `lru_cache(maxsize=8)` on the regex builder avoids recompiling per-entry across a full rescan; no new external input surface.

Osiris-DevWorks (dismissed):
Thanks for the fix. I grepped all call sites of `has_bp_section` / `_bp_section_span` / `extract_bp_item_names` / `apply_owned_to_value`: every one is covered, including the easy-to-miss `apply_to_game` write path, and `_bp_header()` keeps the settings key in one place. The qualifier fix for CIG's "(Repeat Only)"-style headers is pinned by the fixture parity test.

**Note:** the EM-wrapper anchor is a strictness change for the default path too. A hand-edited `custom_value` containing a bare unwrapped "POTENTIAL BLUEPRINTS" used to match the BP Descriptions filter and now won't (one existing test needed the wrapper added). Stock CIG data and generator output always wrap, so no action needed; just worth a release-note line if we think anyone hand-types these.

Osiris-DevWorks (approved):
## Review — #353 renamed blueprints header

Approving. Well-executed fix, and the self-review caught the one change that would have made this worse than the bug it fixes. I verified that fix independently rather than taking the numbers on trust.

**Verification:** full suite **1629 passed, 21 skipped, 0 failed**. Parity and rejection behaviour checked by executing the branch against the shipped fixture:

```
bare-substring matches : 292          qualifier shapes CIG actually ships:
EM-anchored matches    : 292            272  ''
LOST by anchoring      : 0               10  '(Repeat Only)'
                                          5  '(BitZeros Only)'
rejection (header renamed to "stuff"):    3  '(Nyx Only)'
  reject '<EM3>stuffed animals</EM3>'     1  '(Yormandi Eye Only)'
  reject 'Wallace had some stuff...'      1  '(Pyro IV/V Area Only)'
  MATCH  '<EM3>stuff</EM3>'
  MATCH  '<EM4>stuff (Repeat Only)</EM4>'
  reject '<EM3>stuff</EM4>'   <- backreference works
```

Every real qualifier is a single parenthesised run, so the one optional bounded group is exactly the right shape for the data.

**Acceptance against #353**, end to end:

| Criterion | Result |
|---|---|
| Renamed header populates the tracker | OK - `bp_header="stuff"` finds `Antium Core`, `Norfield` |
| Prose collision produces no fake items | OK - `- Ed`/`- H`/`- Wallace` all absent |
| Un-renamed installs unaffected | OK - default path unchanged |
| CIG-qualified headers still work | OK - `Potential Blueprints (Repeat Only)` |
| All four matchers threaded | OK - metadata, owned re-weave, BP filter, apply-to-game |

I also checked a gap that isn't one: `MULTIPLE BLUEPRINT POOLS` isn't threaded through the rename path, but the renameable keys are `details`/`blueprints`/`items`/`blueprint_data` only, and the pools header appears nowhere in the generator. Correctly left alone.

---

### Minor

**1. `_bp_header()`'s docstring describes a `None` return that can't happen.** `main_window.py:5308` says *"or None when it has never been renamed"*, but `get_mission_headers()` always falls back to `MISSION_HEADER_DEFAULTS`:

```python
"blueprints": s.value(AppSettings.MISSION_HEADER_BLUEPRINTS, d["blueprints"]),
```

So a fresh install returns `"POTENTIAL BLUEPRINTS"`, and `.get()` can't miss since the key is unconditionally populated. The code is right either way — the dedup guard folds the default string and its variants onto the default-only pattern:

```
bp_header=None                        -> identical to None-pattern? True
bp_header='POTENTIAL BLUEPRINTS'      -> True
bp_header='potential blueprints'      -> True
bp_header='  POTENTIAL BLUEPRINTS  '  -> True
bp_header='stuff'                     -> False
```

Only the docstring needs correcting; it currently implies passing the default string is a bug. The `"str | None"` annotations downstream stay right.

**2. `entry_filter.py`'s change has no test.** `tests/test_entry_filter.py` has no `bp_header` coverage, though it already carries three `bp_descs_only` tests to model one on. Worth closing because the filter and the tracker are independent matchers over the same data — if they drift, a row shows under "BP Descriptions" but never reaches the tracker, which is an annoying bug to chase. (The existing entry-filter tests do still pass under the anchoring; `test_bp_descs_only_recognises_multiple_blueprint_pools_header` already uses a wrapped header.)

**Nitpicks:** nothing pins the empty/whitespace-`bp_header` fallback, and `_bp_section_span`'s claim that a renamed *end-boundary* header is already handled by the generic EM scan is untested.

---

### Worth calling out

`test_fixture_header_recognition_matches_bare_substring_search` is the right response to a bug that every hand-written case missed — when your fixtures share a blind spot with your implementation, only real data breaks the tie. Worth reusing as a pattern. The six parametrized cases using real CIG header strings, and the comment recording *why* the earlier tests passed through the regression, are the other half of that.

`test_qualifier_tolerance_does_not_admit_unrelated_wrapped_text` is the necessary counterweight — widening a regex without a matching tightening test is how the next regression gets in.

The one modified existing test (`"POTENTIAL BLUEPRINTS"` -> `"<EM3>POTENTIAL BLUEPRINTS</EM3>"`) is a tightening to match reality, not a loosening to accommodate a bug, and it's the only one. Correct call.

`_bp_header()` is also exactly the accessor pattern worth applying to `enclosings_from_tag_configs(AppSettings.get_all_tag_configs())` in #355, where it's still repeated six times — hoisted out of both 90k-entry loops here, with the read-fresh-don't-cache decision reasoned out.

---

### Merge order

This and #355 both add a different keyword to `extract_bp_item_names`, `apply_owned_to_value`, `build_blueprint_metadata` and `filter_entry_indices` — whichever merges second conflicts in four files. The parameters are orthogonal; keep both.

Suggest **merging this one first**. It's approvable now, whereas #355 needs a fix to `parse_component_tag`'s heuristic gate and will be touched again regardless — better it absorbs the conflict resolution than that this waits on it.

Issue #357, opened 2026-08-07
Add a delete all saved settings tickbox to uninstaller


Stealrull:
As a  discord user pointed out it's troublesome to do a fresh install of smart citizen if something has gone wrong as settings are left over on uninstall, adding a tickbox to delete all user settings onn uninstall would fix this in the future for anyone wanting to do a full reset.

Would need to be part of the uninstall process and something the user has to manually tick if they want to remove everything, along with a warning when they tick it that this will remove all saved settings.

Should remove these three things:

HKEY_CURRENT_USER\Software\Osiris DevWorks\Smart Citizen 
Documents\Smart Citizen\ (or wherever user_data_dir pointed, if customised)
%LOCALAPPDATA%\Smart Citizen\ (or wherever CACHE_DIR pointed, if customised)

Stealrull:
Fixed in PR #384

PR #359, opened 2026-08-08
Add opt-in "delete all saved settings" checkbox to the uninstaller (#357)

Stealrull:
Closed, superseded by #384 -- same fix for #357, moved to a branch on this repo instead of the Stealrull fork so `claude-review` can run on it. Kept here for the review discussion below.

## Summary
- Uninstalling has always deliberately left the registry node, user data folder, and DataForge cache behind (the "#172 persistence lock") so a reinstall picks up where the user left off. That's the right default, but leaves no way to get a genuine clean slate when something has gone wrong -- a Discord user reported exactly this friction.
- Adds a custom uninstall confirmation dialog with an unchecked-by-default "Also delete all saved settings" checkbox. Ticking it pops up a separate warning listing exactly what gets deleted, requiring explicit OK to proceed -- declining reverts the checkbox rather than silently continuing checked.
- Only when explicitly confirmed does the uninstaller delete all three locations named in the issue:
  1. `HKEY_CURRENT_USER\Software\Osiris DevWorks\Smart Citizen`
  2. The user data folder -- override-aware (`user_data_dir`), not just the `Documents\Smart Citizen` default
  3. The separate DataForge cache folder -- also override-aware (`cache_dir`); this one was never cleaned by the normal uninstall path at all, custom location or not, so this closes a second, smaller pre-existing gap too
- Leaving the box unchecked is byte-for-byte the existing uninstall behavior. The `#172` persistence-lock comment block is preserved and extended to document this as its one deliberate, explicitly-gated exception.
- The dialog is skipped entirely during a silent uninstall (`UninstallSilent()`) -- the in-app auto-updater's "remove old version" step runs its uninstaller with `/VERYSILENT`, and a raw custom VCL form isn't auto-suppressed by Inno's silent flags the way `MsgBox` is. Without this guard an automated upgrade would hang forever waiting for a click nobody will make.

## Test plan
- [x] `ISCC.exe installer.iss` compiles clean (no Pascal Script errors) -- this feature has no Python code, so the compiler is the only automated check available.
- [x] Manual test, two rounds: first round found the checkbox caption running off the dialog (`TNewCheckBox` doesn't word-wrap like a text label) and asked for the warning to be conditional/require explicit acknowledgment rather than static always-visible text. Second round (after moving the warning into a popup fired on check, with revert-on-cancel) confirmed working as intended.

### DIFF
--- a/installer.iss
+++ b/installer.iss
@@ -894,9 +894,10 @@ begin
   { Skip if it's the same folder as (or nested under) the user data folder
     already just deleted above -- a custom cache_dir is normally a wholly
     separate location, but avoid a redundant/confusing second DelTree pass
-    over the same path either way. }
-  if (CacheDir <> '') and
-     (CompareText(RemoveBackslash(CacheDir), RemoveBackslash(UserDataDir)) <> 0) then
+    over the same path either way. PathUnderRoot covers both cases: it
+    matches an exact match too, since normalizing both sides the same way
+    before the prefix test makes a path trivially "under" itself. }
+  if (CacheDir <> '') and (not PathUnderRoot(CacheDir, UserDataDir)) then
   begin
     if DirExists(CacheDir) then
     begin
--- a/installer.iss
+++ b/installer.iss
@@ -92,6 +92,12 @@ var
     itself falls back to displaying English (index 0) for an unrecognized
     value. See #236 review discussion. }
   LanguageChoiceUnknownSaved: Boolean;
+  { #357: set by the checkbox on the custom uninstall confirmation dialog
+    (InitializeUninstall). False (the safe, non-destructive default) unless
+    the user explicitly ticked "also delete all saved settings" -- read by
+    CurUninstallStepChanged to decide between the normal cache-only cleanup
+    and the full opt-in wipe (DeleteAllUserSettings). }
+  DeleteAllSettingsChecked: Boolean;
 
 function IsAutoUpdate(): Boolean;
 begin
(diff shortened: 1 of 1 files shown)


Stealrull (commented):
## Review

**Overview:** Adds a custom uninstall confirmation dialog with an unchecked-by-default checkbox. Ticking it triggers a separate acknowledge-to-proceed warning popup (declining reverts the checkbox); only on explicit confirmation does the uninstaller wipe the three locations the app's settings/data can live in: the HKCU registry node, the user data folder, and the separate DataForge cache folder -- the latter two resolved override-aware (respecting `user_data_dir`/`cache_dir` if the user relocated them), not just their defaults. Leaving the box unchecked preserves today's exact uninstall behavior.

**Code quality:** Careful, deliberate handling of the pre-existing "#172 persistence lock" invariant -- rather than bypassing or weakening that documented guard, the PR extends its own comment to explicitly name this as its one gated exception, which is exactly the right way to touch code with a "do not reintroduce X" warning already attached to it. Correct, non-obvious ordering: both override paths (`user_data_dir`/`cache_dir`) are resolved from the registry *before* the registry node is deleted -- getting this backwards would silently fall back to default paths and orphan a custom-location folder; the comment calls this out explicitly. Good defensive UX: the checkbox's `OnClick` handler reverts to unchecked on anything other than an explicit OK, a fail-safe default rather than fail-open. The `UninstallSilent()` guard against a hung automated upgrade is a real, non-obvious correctness concern (unlike `MsgBox`, custom Inno forms aren't auto-suppressed by silent flags) -- good catch, well-documented.

**Minor suggestion (non-blocking):** The "skip re-deleting the cache folder if it's the same as the user data folder" check only catches an exact match, not a `cache_dir` nested *under* `user_data_dir`. This file already has a `PathUnderRoot` helper built for exactly this kind of prefix check (used for OneDrive detection) that could be reused here. Not a correctness bug though -- the nested case just falls through to a `DirExists` check on an already-deleted path and logs "absent, nothing to delete" harmlessly. Worth tidying if touching this code again.

**Test coverage:** Appropriately scoped and honestly caveated -- this is pure Inno Setup Pascal Script with no unit-testing framework available, so `ISCC.exe` compiling clean plus manual testing is the realistic ceiling. The PR description documents that manual testing caught two real issues (checkbox caption overflow, warning needing to be conditional/acknowledged) fixed before this version -- good signal the testing was substantive.

**Performance/security:** No concerns. All registry/folder operations are scoped to per-user locations -- running under `PrivilegesRequired=admin` doesn't expand the blast radius beyond what a normal user could already touch themselves. Registry deletion is correctly scoped to the app's own specific subkey, not the shared publisher key, so no risk to sibling apps.

PR #362, opened 2026-08-12
Fix Yormandi Eye and Irradiated Valakkar Pearl blueprint lists merging together

### LINKED ISSUE #361: Yormandi Eye collection missions blueprint lists issue
As reported by JakP on discord these missions list items not correctly due to sharing the same names.

Image

Easy fix to separate them.

Stealrull:
## Summary
- A Discord user (JakP) reported that Rayari research missions ("Additional Resources For Research") show blueprint items from two unrelated reward sets mixed into one list, with third-party trackers only ever showing one set.
- Root cause: multiple mission variants (one requiring an Irradiated Valakkar Pearl, another a Yormandi Eye) share the same title AND description text, so their separate blueprint reward pools were being flattened together during generation.
- `_merge_blueprint_pool` now keys pools by their own identity instead of by label alone, so genuinely distinct pools stay distinct instead of collapsing into one list when they share an empty (non-rank) label.
- Since CIG reuses the same description across every variant of this mission (confirmed via cross-referencing SCMDB), there's no way to show only the relevant pool per mission instance from data available in the pipeline. Instead, the two known pools are labeled by what they actually reward (`[Yormandi Eyes]` / `[Irradiated Valakkar Pearls]`) via a small hand-curated lookup, so a player can tell them apart regardless of which specific mission instance they're viewing.
- Section order is deterministic: sorted by header text and pool content, not by whichever contract XML happens to get scanned first, so the two sections render in a stable order across regenerations.
- The label overrides match on exact item-set equality. If a future CIG patch changes either pool's items, the override will stop matching — a warning is now logged when a pool closely resembles a known override without matching it exactly, so the table gets flagged for updating instead of silently going stale.

## Test plan
- [x] `pytest -q` — 1591 passed, 21 skipped
- [x] New regression tests covering: distinct pools staying separate, identical pools still deduping, colliding headers getting disambiguated, the manual label override producing the exact requested text, section order staying stable regardless of insertion order, and the drift-warning firing on a near-miss while staying silent on exact matches and unrelated pools
- [x] Built a portable and manually verified the fix against real extracted game data (Rayari research missions now show `[Yormandi Eyes]` and `[Irradiated Valakkar Pearls]` instead of one merged list)

Fixes #361

### DIFF
--- a/scripts/generate_enhancements_ini.py
+++ b/scripts/generate_enhancements_ini.py
@@ -3151,7 +3151,7 @@ def _merge_blueprint_pool(
 # plumbing is needed -- just what's already visible in the rendered list.
 # Add entries here as more ambiguous multi-pool missions are found; any
 # pool not listed keeps the automatic "first item" naming below.
-_BLUEPRINT_POOL_LABEL_OVERRIDES: dict[frozenset, str] = {
+_BLUEPRINT_POOL_LABEL_OVERRIDES: dict[frozenset[str], str] = {
     frozenset({
         "P8-AR Rifle", "P8-AR Rifle Magazine (15 Cap)",
         "Palatino Arms", "Palatino Arms Moonfall",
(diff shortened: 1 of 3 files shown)


Stealrull (commented):
## Overview

Same core fix as before (issue #361): blueprint pools for two unrelated Rayari research mission variants no longer collapse into one undifferentiated list. This revision adds two follow-ups from the earlier review round:
1. **Deterministic section ordering** — sections are now sorted by `(system/label keys, header text, fingerprint)`, so which of `[Yormandi Eyes]` / `[Irradiated Valakkar Pearls]` renders first no longer depends on filesystem/XML-scan order. Verified with a new test that feeds the two pools in reversed insertion order and asserts identical output.
2. **Label wording fix** — `"Yormandi Eye's"` / `"Irradiated Valakkar Pearl's"` -> `"Yormandi Eyes"` / `"Irradiated Valakkar Pearls"` (dropped the incorrect possessive apostrophes).

## What's resolved from the last review pass

- The non-deterministic ordering flagged previously is fixed and specifically tested (`test_override_order_is_independent_of_insertion_order`).
- The fix is more thorough than strictly necessary — it also makes the *non-override* collision path (Pass 1's "first item" naming) order-deterministic via the same fingerprint tiebreak, not just the override case. Good defensive completeness.

## Remaining minor items

1. **Override table is exact-match/silent-degrade.** If CIG adds, removes, or renames an item in either pool in a future patch, `frozenset` equality stops matching and the section quietly falls back to auto-generated naming with no warning. Acceptable given the hand-curated, "add entries as found" design, just worth remembering as a maintenance point.
2. **Minor type-hint nit**: `dict[frozenset, str]` could be `dict[frozenset[str], str]` for precision.

Neither is blocking. Test coverage is strong — 51 tests in `test_blueprint_pools.py`, all passing, directly covering the new ordering guarantee.

Stealrull (commented):
## Follow-up: both previously-flagged minor items are now resolved

1. **Type-hint precision** — `dict[frozenset, str]` -> `dict[frozenset[str], str]`. Done.
2. **Silent override drift** — added `_warn_if_near_override_miss`: if a pool shares >=70% of its items with a known override without exactly matching (e.g. a future CIG patch adds/removes/renames an item), it now logs a warning naming which override it resembles, instead of quietly reverting to auto-generated naming with no signal. Threshold is conservative to avoid false positives on unrelated pools that happen to share a common battery/magazine item. Covered by 3 new tests: near-miss fires the warning, an exact match doesn't self-warn, and a genuinely unrelated pool stays silent.

Verified the new code directly (not just via the test suite): fed a synthetic pool missing one item from the real "Yormandi Eyes" set and confirmed the warning fires with the correct pool/override names in the message; fed the exact real override and confirmed it stays silent.

No new issues found in this pass. 54 tests in `test_blueprint_pools.py`, all passing; 1591 total across the suite.

Osiris-DevWorks (approved):
Thanks for the fix. I verified the language gate independently: `en_loc = loc` is the same object on an English run and a separately parsed dict otherwise, and `tag_loc` is always present in ctx, so the identity check is a correct signal. Per-contract grouping preserves the Adagio same-contract merge while separating cross-variant pools, and the sub-contract path correctly reuses the parent's pool data. Section ordering is deterministic. Test coverage here is the strongest of the six 2.3.1 PRs.

**Note (your call):** the drift warning only reaches the log. Since the override table is the load-bearing part of the fix, consider a tester Test Plan line ("Rayari research missions still show both labeled sets") so a CIG data change gets noticed without anyone reading generation logs.

PR #365, opened 2026-08-14
[#364] Add horizontal scrollbars, free window resize and resizable String Editor columns

### LINKED ISSUE #364: Add missing horizontal scroll bar
As mentioned on discord if the app isn't in full screen some windows need a horizontal scroll bar which is missing, when adding this in testing I found that with that we can also have the app be able to be resized freely if the user desires, this also opens up string editor columns to be able to be resized as desired as well.

I specifically added a reset window proportions option in the more menu as well with a popup warning just in case.

Stealrull:
Fixes #364

Below full screen some tabs had no horizontal scrollbar, so content was either squeezed or clipped with no way to reach it. Fixing that also made free window resizing and user-resizable String Editor columns possible, as described in the issue.

## Consistent scrollbars

Every tab's bars now sit in the same place. The String Editor was the only content tab without a scroll area of its own, so its overflow fell through to the window-level one, which renders *below* the tab widget: 70px from where Config and Enhancements put theirs.

Page margins are zeroed so each tab's own bar sits flush at the tab edge, and the "Showing N of M" / "Lines: N" labels moved above their views, since a widget below a view stops that view's bar reaching the bottom edge. That alone was leaving the Log tab 24px out of line.

Measured at 900px: every tab-bottom horizontal bar at `y=599`, every vertical bar at `x=876`, no window-level bar, nothing offscreen, and a single window size across an 8-tab sweep, so no tab-switch resize (the regression behind #65).

## Resizable columns

Columns are `Interactive`, the only mode Qt lets the user drag and the only one it auto-fits on a divider double-click (a 150px column snaps to 727px, exactly `sizeHintForColumn`). The default layout is computed to match what `Stretch`/`ResizeToContents` produced, validated against Qt's own output at four viewport widths, and recomputed on every table resize until the user takes over, so a fresh install looks unchanged.

Two header bugs had to be fixed for this to work at all:

* A drag was swallowed the moment the pointer left the header's label strip, and that strip is only **16px** tall, so it was the common case rather than an edge case.
* The filter boxes did not follow a horizontal scroll (column at `x=-1115`, box stranded at `x=118`). Unreachable while the columns stretched and never overflowed; making them resizable is what made it reachable.

## Persistence

Window geometry and column widths persist. Widths are written only once the user actually changes one, so absence keeps meaning "never customised" and every fresh install gets defaults fitted to its own screen.

A **Reset Window Proportions** action in the More menu restores geometry, dock layout and column widths, behind a confirmation stating that settings and localization data are untouched. Its strings are in all 8 locales, in `at` with `ht` empty for the translated ones per the activation tests.

## Also fixed

* **The filter row covered the column names after any reload.** The header's label height reports 0 during transient layout passes, and a model reset is one, so the editors were placed at `y=0` over the labels. Only a section resize recovered it, so a reload producing identical widths left them there, which is why it showed on first-run Generate Enhancements. The height is now cached, including for the label painting, which had the same read and a documented history of the header text vanishing until restart.
* **Horizontal scrolling jumped a whole column per step** (300px, only 7 scrollbar positions across 1734px of content). Now per-pixel with a fixed wheel step.
* **Simple mode opened with a scrollbar** over a page with room to spare, and wasted 109px above the footer links. The two views were pages of a `QStackedWidget`, which always reports its tallest page, so Simple kept reserving height for the Advanced tabs. They are now siblings swapped by visibility, where the hidden one costs nothing. The window went from 685x506 to 685x397, and the branding labels no longer stretch apart on a maximised window.

## Notes for review

**Two structural choices worth flagging:**

The Simple/Advanced switch no longer uses a `QStackedWidget`. Reserving for the tallest page is inherent to that widget, and working around it needed a size-policy override, a height cap and filler spacers that each broke a different way; plain show/hide removes the cause instead. The tabs and the Simple page are siblings in the content layout with the same stretch.

An earlier attempt at the squish problem pinned each tab page's `minimumWidth`. That propagates into the window's own minimum, inflating the canvas to 2258px inside a 900px window and pushing each tab's bars off the right edge, so they looked absent until you scrolled sideways. It was also redundant, since the only widget it protected already contains its own overflow. A regression test pins `minimumWidth == 0` so it doesn't come back.

## Testing

`pytest`: **1632 passed, 21 skipped**, up from 1587 on the base.

New coverage for the scrollbar placement contract across all tabs, the column default layout and its persistence (corrupt values, wrong-length stored layouts, the `minimumSectionSize` clamp), saved-geometry precedence, the reset action, and the header regressions. Each new guard was confirmed to fail when its fix is reverted, rather than only passing.

`tests/gui_window.py` holds the shared setup for the two tests that build a real `MainWindow`, documenting the three ways doing so can destabilise a pytest run (a background update-check thread, the root-logger error dialog handler, and destroying the window mid-session). The lightweight stub in `test_ui_mode.py` is still preferred wherever it can observe the behaviour.

Verified in a portable build on Windows: `SmartCitizen-Portable-v2.3.1`.

---

**Update (27 Aug):** both items from your review are addressed.

- `string_column_widths` is in `PROFILE_EXCLUDE_KEYS` (`491937a`), with a test naming the key explicitly, so an exported settings zip no longer carries a screen-specific column layout.
- The HELP.md docs gap is filled in `3d9e6c5`, across all 8 languages: the resizable-columns bullet, a new Window Layout section covering **Reset Window Proportions**, and section 13's exclusions list now naming the column widths. Logged in `languages/TRANSLATIONS.md` as an AI docs-parity pass.

Your caching note (`_apply_default_column_layout` re-reading saved widths on every viewport resize) is the one item left open, which you marked low priority -- happy to take it here or defer it.


Osiris-DevWorks (commented):
Thanks for the fix. This is the most carefully measured of the six, and the risky structural pieces check out: no stale `view_stack` references anywhere, both siblings keep stretch=1, the mouse-forwarding fix self-heals a lost release, `_label_row_height` caches around Qt's transient zero-height passes, and the i18n keys are complete in all 8 locales. One real fix before merge, plus two smaller items.

**Fix: add `string_column_widths` to `PROFILE_EXCLUDE_KEYS`** (`src/utils/settings.py`, next to `window_geometry` / `window_state`). That set strips machine-specific layout from Export Settings, and HELP.md promises exactly that, but the new column-widths key will now travel in an exported settings zip and get adopted verbatim on a machine with a different screen. One line plus a test entry.

**Note:** `_apply_default_column_layout` runs on every table-viewport resize and calls `AppSettings.get_string_column_widths()` (a settings read plus JSON parse) each time until the user takes over the columns. The section hints are cached; the saved-widths read could be too. Low priority.

**Note:** Reset Window Proportions and the resizable columns aren't in `docs/HELP.md` yet. The pre-release docs sync would catch it, but it is cheaper to add now while the details are fresh.

Stealrull (commented):

Reviewed the full diff against Smart Citizen's layer ordering and standards, alongside the review already on this PR.

**Blocking item — resolved.** Osiris's one required fix (`string_column_widths` missing from `PROFILE_EXCLUDE_KEYS`, so machine-local column widths would travel in an exported settings zip and be adopted verbatim on a different screen) is applied, with a test naming the key explicitly and a comment tying it to the same rationale as `window_geometry` / `window_state`.

**His two Notes** remain open and are both marked low-priority/optional by him: caching the saved-widths read in `_apply_default_column_layout`, and adding Reset Window Proportions + resizable columns to `docs/HELP.md`. The docs one would otherwise be caught by the pre-release docs sync.

**Fresh pass over the risky structural pieces:**
- `COL_*` constants used throughout the new column-mode mapping -- no hardcoded indices introduced.
- `_sync_scroll_connection`'s `RuntimeError`/`TypeError` guard is correct and well-justified: `setHorizontalScrollBar` destroys the bar it replaces, so a stale wrapper genuinely does raise on disconnect.
- The `_forwarding_mouse` self-heal (clearing on a no-button move) closes the lost-release case that would otherwise leave every later hover forwarding.
- `_label_row_height`'s cache is applied consistently across all four read sites (`paintSection`, `paintEvent`, `_position_editors`, and the three mouse handlers), so no path still reads the raw collapsing `sizeHint`.
- i18n: all 8 locales carry the 4 new keys, with `ht` empty on translated locales per the activation tests.

Nothing new to add. No blocking issues from this pass.

Osiris-DevWorks (approved):
## Review — #364 scrollbars, free resize, resizable columns

Approving. Large but coherent, and the parts that could have gone wrong quietly are the parts that are best guarded. Two Minors, neither blocking.

Given your own caveat that this is the PR where your review is weakest, I ran and measured rather than reading:

| Check | Result |
|---|---|
| Full suite | **1632 passed, 21 skipped, 0 failed** |
| New GUI test files | 12/12 **run, no skips** under offscreen — the scrollbar contract is genuinely guarded, not silently skipped |
| i18n coverage | **0 problems** across 8 locales x 5 new keys, `ht`/`at` convention correct in each |
| per-pixel scroll cost | `_position_editors` = 5.1us; a full 1734px scroll ~= **9ms** of repositioning |
| `retire_main_window` | one repeating timer survives (finding 2) |

Two suspicions I disproved, recorded so nobody re-raises them:

- **`QSettings.Format.IniFormat` in `tests/gui_window.py`.** `src/utils/CLAUDE.md` says IniFormat "mangles slash-prefixed keys ... and can't round-trip them", so I expected a trap for a future MainWindow test touching `tag_builder/{category}/config`. It round-trips fine — that key and `enhancements/categories/ships/enabled` both came back intact. Not a finding.
- **Per-pixel scrolling x editor repositioning.** Going per-pixel takes `valueChanged` from 7 emissions to ~1734, each repositioning every filter editor. 9ms across a full scroll. Not a finding.

My first attempt to measure scrollbar alignment reported two distinct rows and `minimumWidth = 68`, which looked like a contradiction. It wasn't — I was counting nested scroll areas and `SingleRowScrollArea`, and checking the *window's* minimumWidth rather than the tab page's. `test_tab_scrollbar_placement.py` excludes both, correctly, and the guard is on `stub.tabs.widget(0)`. Your methodology is right; mine was naive.

**Release scope:** you flagged this as a possible 2.3.1-vs-2.4.0 call. I think it's already settled — #364 is yours, carries the `2.3.1` label, and explicitly scopes in all three parts including the reset action ("I specifically added a reset window proportions option in the more menu as well with a popup warning just in case"). Not raising it.

---

### Minor 1 — three `Experiment:` comments in shipping code

`main_window.py`, at the `setMinimumSize(0, 0)` call, the central-widget `QScrollArea`, and the filter-row wrapper. All three are load-bearing, and the second says so in its own body:

```
# Experiment: route the content through a QScrollArea instead of
# ...
# Removing it is not an option: without it the window hard-floors around 666px wide
```

A comment that calls itself an experiment while stating it can't be removed sends the wrong signal — the next maintainer reads "provisional, safe to rip out". The prose underneath is good; only the lead-in needs to change (`# Route the content through a QScrollArea ...`).

### Minor 2 — `retire_main_window` leaves a repeating timer running for the rest of the session

Never destroying the window is the right call given the documented 1-in-4 access violation. But retirement only hides it and detaches the log handler. Measured:

```
active QTimers while shown : 1
active QTimers after retire: 1
   still running: interval=300ms singleShot=False parent=MainWindow
```

That's `_eye_pulse_monitor` (`main_window.py:1094`), firing every 300ms into a hidden-but-alive window for the rest of the run, once per window ever built. Harmless at two consumers; a plausible flake source as more MainWindow tests land — which is the growth this module exists to support. Two lines:

```python
def retire_main_window(window) -> None:
    ...
    for timer in window.findChildren(QTimer):
        timer.stop()
    window.hide()
```

Worth doing precisely because the docstring is trying to stop the next author rediscovering this class of problem.

---

### Worth calling out

**`test_every_tab_puts_its_horizontal_bar_on_the_same_line` has a vacuity guard** — `assert len(contributors) >= 2` before asserting they share a `y`. Without it the test passes trivially whenever only one tab overflows. That's the difference between a test and a decoration, and it's rare to see written down.

**`_is_nested` documents a CI failure of an earlier version of the test itself**, so the exclusion reads as a fix rather than an unexplained carve-out.

**`_section_size_hints` caching is real performance work.** `sectionSizeHint` walks every viewport row through the delegates (~270 calls across nine columns), and this ran per frame of a window drag before the cache. The `changeEvent` invalidation plus the `retranslate_ui` / `_on_loading_finished` call sites cover the ways hints actually move.

**Replacing `QStackedWidget` with sibling show/hide deletes complexity** — the `Ignored`/`Expanding` policy dance goes away, and it fixes the cause (a stack reports its tallest page) rather than compensating for it.

**Persistence is properly defensive**: corrupt widths degrade to `[]`, a stale list whose length no longer matches the column count is ignored rather than applied piecemeal, the save is gated on `_user_resized_columns`, and `reset_window_layout` removes keys rather than writing defaults so absence keeps meaning "never customised".

`_suppress_column_capture` correctly separates programmatic sizing from a real drag. `_forwarding_mouse` self-heals on a buttonless move. `_sync_scroll_connection`'s `RuntimeError` catch is the right one — `setHorizontalScrollBar` destroys the bar it replaces. And the `GetWindowPlacement` / `rcNormalPosition` note explains an obscure Windows behaviour that would otherwise look like an arbitrary pre-maximize `resize()`.

---

### Action items

- [ ] Drop the `Experiment:` lead-in from the three comments
- [ ] Stop timers in `retire_main_window`

### Merge order

No overlap with #355/#358 — this doesn't touch `owned_items.py`, `blueprint_meta.py`, or `entry_filter.py`'s matching functions. Can merge independently of that pair.

PR #366, opened 2026-08-16
Run the Tag Builder freshness check at startup, not only on channel switch (#363)

### LINKED ISSUE #363: Bug - Modification of blueprint rewards to allow tags
Would it be possible, to provide a feature, allowing modification of the Blueprint rewards on missions, if they are componets, to show tags, Ex [M-A-S2] right in the mission rewards of the mission details?

Stealrull:
Fixes #363.

## The report

The "annotate component tags in mission descriptions" checkbox read enabled, but the annotations were absent in game. Erelaszun found that turning it off, letting the generation finish, then turning it back on fixed it, and guessed that the checkbox "doesn't register as enabled until its cycled".

The setting always registered. What was never checked is whether the generated INIs on disk were built from it.

## Root cause

Three staleness signals exist. Only two of them were consulted at launch:

| Signal | At startup | On channel switch |
| --- | --- | --- |
| DataForge cache vs `Data.p4k` | yes | yes |
| enhancement output files missing | yes | yes |
| **tag config vs the config the INIs were generated from** | **no** | yes |

The annotate toggle lives entirely in that third row. `refresh_tag_builder_dirty_state()` implements it correctly, lighting Save Tag Changes when the recorded stamp is missing or differs from the live config, but it was wired to exactly one caller: `_on_channel_changed`. At launch, `setup_ui` ended with an unconditional `_set_tag_btn_dirty(False)`.

So the button was grey at startup regardless of whether the files on disk matched the setting. Generate Enhancements was grey too, because `_compute_initial_enhancements_dirty()` only asks whether the output file exists, never whether it was built from this config. Two grey buttons over stale output, and nothing saying anything was wrong.

Cycling the checkbox worked because `_mark_tag_dirty` is a blind "the user touched a control" flag rather than a comparison. The config fingerprint is identical before and after an off/on cycle; the cycle simply forged the dirty state the startup check should have derived.

Reproduced with a real window against a seeded profile (generation complete, stamp from an older config):

```
annotate checkbox ticked        : True
output is stale per the app     : True
Generate Enhancements enabled   : False
Save Tag Changes enabled        : False     <- no way to act
```

Every install upgraded from before the stamp existed has no stamp at all, since the stamp and the annotate flag joining the fingerprint landed in the same commit. This is not an edge case, it is what upgrading users get.

## What changed

Every point that can invalidate the generated INIs now re-derives the button instead of asserting a state:

- **Launch** (`setup_ui`), the reported bug.
- **Import Settings** (`reload_tag_builder_from_settings`), which cleared the button after an import that changed the config, leaving the INIs built from the pre-import one.
- **Export Settings** (`flush_pending_tag_edits`), which persists on-screen edits without regenerating, and threw away a red button the user had already been shown.
- **A finished run**, since both trigger paths clear the button the moment they launch one. A run that then failed left it grey over INIs the tag config never reached; the stamp is only written on success, so re-deriving relights it for a retry.
- **A language switch**, which recomputed neither freshness button even though the INIs and the stamp are both per-language. `_on_channel_changed` already made this exact pair of calls, language switching had simply never been wired up to match.

Fixing only the launch case would have left the same bug reachable four other ways, which is why the change is wider than the report.

One behaviour bound comes with it: being stale requires something that can be stale, so the check reports clean when no enhancement output exists at all. Otherwise a brand new install, which has no stamp and therefore reads as unknowable, would show a red Save Tag Changes before the user had done anything, alongside the Generate Enhancements button already lit for the missing files.

## Testing

`tests/test_tag_config_staleness_startup.py` (new) covers the startup states, the annotate toggle in isolation, the no-output gate, Export Settings, the language switch, and both generation outcomes. Nine assertions across the two test files fail against the unfixed source, so none of them are vacuous. Full suite: 1587 passed, 21 skipped.

`tests/test_tag_builder_import_refresh.py`'s fixture now redirects the user data dir as well as the settings backend. Swapping the backend alone left `get_enhancements_dir()` pointing at the developer's real install, which is where the stamp is read from, so stamp assertions would have passed or failed depending on the machine. Its `test_refresh_leaves_save_button_clean` is renamed and seeds a matching stamp, because "a refresh always clears the button" was the behaviour being fixed. The guarantee it protects, that a refresh must not light the button for unsaved edits, is still pinned.

Verified in a portable build: a fresh unzip leaves Save Tag Changes grey and only Generate Enhancements lit, and the Import Settings route above turns Save Tag Changes red once there is output to be stale.

## Note for merging: #358 conflicts with #355

This PR touches neither, but the integration build used to test it had to exclude #358, and the reason is worth flagging before either is merged.

#355 (`enclosings=`) and #358 (`bp_header=`) each add a **different** new keyword parameter to an overlapping set of functions:

| Function | #355 adds | #358 adds |
| --- | --- | --- |
| `owned_items.extract_bp_item_names` | `enclosings=` | `bp_header=` |
| `owned_items.apply_owned_to_value` | `enclosings=` | `bp_header=` |
| `blueprint_meta.build_blueprint_metadata` | `enclosings=` | `bp_header=` |
| `entry_filter.filter_entry_indices` | body only | `bp_header=` |

Whichever merges second will conflict in `src/gui/main_window.py`, `src/utils/blueprint_meta.py`, `src/utils/entry_filter.py` and `src/utils/owned_items.py`.

The resolution is mechanical rather than a design decision. The two parameters are orthogonal, one being the bracket style used to strip a tag and the other a renamed "blueprints" section header, so both should be kept and threaded through together. `#355`'s `enclosings_from_tag_configs()` and `#358`'s `_build_bp_header_re()` both survive unchanged.

### DIFF
--- a/src/gui/enhancements_tab.py
+++ b/src/gui/enhancements_tab.py
@@ -506,12 +506,45 @@ class EnhancementsTab(QWidget):
         genuinely lack those tags); leave it grey when they already match. A
         missing stamp (channels generated before this feature, or never
         generated) reads as dirty, so the button stays clickable exactly as it
-        did before until the next generation records a stamp."""
+        did before until the next generation records a stamp.
+
+        #363 made this the single definition of "the generated INIs don't match
+        the tag config", called from every point that can invalidate them
+        (launch, an import, a finished run) rather than only a channel switch,
+        so it also gained the gate below: being stale requires there to be
+        something that *can* be stale. With no enhancement output on disk there
+        is nothing to be out of date, and Generate Enhancements is already
+        lit for the missing files, so lighting this one too would just put a
+        second red button in front of a new user who has done nothing wrong.
+        """
         if getattr(self, "_apply_tag_btn", None) is None:
+            return
+        if not self._enhancements_output_exists():
+            self._set_tag_btn_dirty(False)
             return
         stamp = AppSettings.get_tag_config_stamp()
         current = self._live_tag_config_fingerprint()
         self._set_tag_btn_dirty(not stamp or stamp != current)
+
+    def _enhancements_output_exists(self) -> bool:
+        """True if any enabled category has at least one generated INI on disk.
+
+        Deliberately "any", not "all": a partially generated set is still
+        output that a tag-config change can invalidate, and the missing half is
+        Generate Enhancements' business (_compute_initial_enhancements_dirty
+        lights it for exactly that). Guarded with getattr so it stays a safe
+        no-op if the enhancements group hasn't been built — setup_ui builds it
+        before the Tag Builder group, so in practice it always has.
+        """
+        checkboxes = getattr(self, "_enhancements_checkboxes", None)
+        if not checkboxes:
+            return False
+        cache_dir = AppSettings.get_cache_dir()
+        return any(
+            (cache_dir / fn).exists()
+            for key, cb in checkboxes.items() if cb.isChecked()
+            for fn in self._files_for_category(key)
+        )
 
     def _on_category_checkbox_changed(self):
         """Enable Apply button if any checkbox differs from saved settings."""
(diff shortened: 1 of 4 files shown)


Osiris-DevWorks (approved):
Thanks for the fix. The re-derive-everywhere approach is right: launch, import, export, generation-finish and language switch all funnel through `refresh_tag_builder_dirty_state()`, and the "nothing generated means nothing can be stale" gate keeps a fresh install from opening with a red button. The per-language path unification (all five checks on `get_enhancements_dir()`) matters most in `_apply_category_changes`, which was renaming English INIs out from under a non-English user. Test coverage is thorough and the fixtures now isolate the user data dir.

**Note:** the re-enable direction of `_apply_category_changes` can still hit "target exists" (regeneration recreated the active file while the category was disabled), which logs and leaves the orphan `.disabled` behind. Harmless today, since the next disable unlinks it first, but a mirrored unlink on the enable path would close it fully.

PR #371, opened 2026-08-19
Pick the live Star Citizen install, not the first one found (#370)

### LINKED ISSUE #370: Smart Citizen keeps reporting *DataForge extraction failed* (Live)
*Synced from Discord: [Smart Citizen keeps reporting *DataForge extraction failed* (Live)](https://discord.com/channels/1438175448420057323/1539278296682991646)*

Hi 🙂 
I have started the portable version of 2.3.0 for the first time and during converting dataforge database it hangs and can the process can not be completed. 
Is there a way to fix that?

Traceback (most recent call last):
  File "src\gui\workers.py", line 502, in run
  File "src\utils\perf.py", line 18, in wrapper
  File "src\utils\pak_extractor.py", line 399, in extract_dataforge
  File "subprocess.py", line 550, in run
  File "subprocess.py", line 1209, in communicate
  File "subprocess.py", line 1630, in _communicate

Stealrull:
Fixes #370.

## The report

A user's DataForge extraction hung for 30 minutes on every run of the 2.3.0 portable, ending in a bare traceback under "DataForge extraction failed". They reported memory sitting at 99% throughout.

Every symptom pointed somewhere other than the cause. Their `build_manifest.id` matched two known-good installs byte for byte, so the game looked current. Verify Files in the RSI launcher changed nothing.

## Root cause

`_scan_common_sc_install_locations` returned its first hit and stopped:

```python
for letter in string.ascii_uppercase:
    for subpath in _COMMON_SC_SUBPATHS:
        if _is_valid_sc_root(candidate):
            return candidate          # first match wins
```

That walk is drive-major over a subpath list whose first entry is `Program Files\Roberts Space Industries\StarCitizen`. The reporter had two installs:

| | Path | Matches subpath | Drive |
| --- | --- | --- | --- |
| Leftover | `D:\Program Files\Roberts Space Industries\StarCitizen` | #1 | D |
| Real | `E:\Roberts Space Industries\StarCitizen` | #3 | E |

D wins, E is never reached. The orphan still had a valid channel folder so it passed validation, and its `Data.p4k` held an older DataForge database, `Game.dcb` at 114 MB where the current build ships `Game2.dcb` at 330 MB. unforge sat on the mismatched database until the timeout.

Verify Files did nothing because the launcher was dutifully verifying the install it manages, which was not the one we were reading.

The comment above `_COMMON_SC_SUBPATHS` cites a real tester install at `E:\Games\Roberts Space Industries\StarCitizen` as the reason those extra patterns exist. The patterns were right. The selection strategy was not.

## What changed

**Newest `Data.p4k` wins.** An abandoned install's archive is frozen at whenever it stopped being patched, while the live one moves with every game update, so recency is the one signal that separates them without asking the launcher where it thinks the game is. Every channel is checked rather than LIVE alone, since a PTU player keeps that channel current while LIVE goes stale. Ties and unreadable timestamps fall back to the original scan order, so a single-install machine behaves exactly as before.

**Every candidate is logged.** The heuristic can still be wrong, and when it is, a log naming the alternatives turns this into a one-line diagnosis instead of the thread it took. One filesystem walk per candidate feeds both the ranking and the log line, so the install we ranked and the date we report cannot disagree.

**The timeout now explains itself.** `DataForgeTimeoutError` carries the database, its size, and the p4k path being read, and points at the install path *before* suggesting Verify Files, because that ordering is what this issue proved correct:

> The database found was Game.dcb (114 MB) at:
> `D:\Program Files\Roberts Space Industries\StarCitizen\LIVE\Data.p4k`
>
> A healthy install converts in well under a minute, so this usually means Smart Citizen is reading the wrong Star Citizen folder. An old or leftover install still on your drive will look valid but hold outdated game files.
>
> Open the RSI Launcher and check the install location it shows. If it is not the path above, correct it in Smart Citizen's Config tab. If it IS the same path, run Verify Files in the launcher to repair the install instead.

It subclasses `RuntimeError` and is recognised by `DataForgeExtractWorker` the way `P4kLockedError` already is, so the blanket handler does not log at ERROR and stack the global dialog on the friendly one.

**unforge's partial output now reaches the log.** `subprocess.run` on Windows kills the child and re-collects its output onto the exception precisely so a timeout is not silent. Nothing caught the exception, so it was discarded, and the one failure that most needed diagnostics produced none.

## Which population each half serves

Worth being explicit, because the two changes do not cover the same people.

**The detection change prevents recurrence, it does not repair.** `get_sc_install_root` returns a saved `SC_INSTALL_ROOT` before it ever reaches the scan, and the scan persists whatever it picked. So anyone who already ran a build that auto-detected the wrong folder has that path written to settings, `_is_valid_sc_root` accepts it (a leftover install genuinely does have a channel folder), and upgrading changes nothing for them. The reporter would still be pointed at the phantom today had they not repointed it by hand.

Re-validating a saved path on every launch was considered and rejected: it would mean walking every drive letter on startup for every user to catch a rare case, and a saved path is often a deliberate choice that should not be second-guessed.

**The timeout message is what serves the already-affected.** It names the folder being read and tells them to check it against the launcher, which is exactly the step that resolved this report. That is why the message leads with the install path rather than Verify Files.

## Measurements

Taken on two independent healthy installs of build 4.9.188.23497, both of which produce a byte-identical `Game2.dcb` at 330,491,142 bytes:

| | Value |
| --- | --- |
| unforge runtime | 40 seconds (budget is 1800) |
| unforge peak RSS | 785 MB |
| unforge stdout | 175 bytes |
| Output | 61,529 XML files, 1.93 GB |

Worth recording because two plausible-sounding theories died against these numbers. `capture_output=True` buffering unforge's stdout is not a memory problem at 175 bytes, and unforge is not a memory hog at 785 MB. `Data.p4k` size is likewise not a signal: the two healthy installs differ by 3.77 GB while producing identical databases.

## Tests

16 new, in two files.

`tests/test_sc_install_scan_picks_live.py` covers the selection: the orphan-beats-live scenario reduced to its essentials, single-candidate behaviour unchanged, scan order breaking ties, a candidate with no readable p4k losing to a real one, newest-across-channels rather than LIVE alone, all candidates logged when ambiguous, no warning when unambiguous, and an unreadable root scoring zero rather than raising.

`tests/test_dataforge_timeout_message.py` covers the error: the recognisable type, `RuntimeError` compatibility, the database and size in the message, the install path appearing *before* Verify Files, partial output reaching the log, bytes output decoded rather than crashed on, and WARNING rather than ERROR level.

Message added to all 8 languages, translated rather than falling back to English, since a user who cannot read the advice cannot act on it.


Osiris-DevWorks (approved):
## Review — #370 pick the live SC install

Approving. Three Minors, none blocking.

**Verification:** full suite **1630 passed, 21 skipped, 0 failed**. i18n clean — `extract.dataforge_timeout` present in all 8 locales with the right `ht`/`at` convention and all three placeholders (`{dcb}`, `{size}`, `{path}`) surviving every translation, 0 problems.

I exercised `_pick_live_sc_install` directly against temp installs with real mtimes rather than reading it:

```
#370 scenario (orphan listed first, as the old scan order would):  E_RSI          OK
Single install unchanged:                                          D_ProgramFiles OK
Tie falls back to scan order (stable sort):                        A_first        OK
Candidate with an unreadable p4k loses to a real one:              E_RSI          OK
```

The root-cause analysis holds up and the fix does what it says.

---

### Minor 1 — the ranking is guarded against bad filesystem timestamps; the log line built from the same numbers is not

`_newest_p4k_mtime` wraps `.stat()` in `try/except (OSError, ValueError)`, so unreadable filesystem data was already on your mind. `_pick_live_sc_install` then formats those same values unguarded:

```python
datetime.datetime.fromtimestamp(mtimes[c]).strftime("%Y-%m-%d") if mtimes[c] else "none"
```

`if mtimes[c]` only catches zero. I assumed Windows would refuse an out-of-range mtime, so I tested it — it doesn't:

```
utime(32503680000)   OK, fromtimestamp OK
utime(64060588800)   ACCEPTED by Windows but fromtimestamp RAISES OSError   <- reachable
utime(253370764800)  ACCEPTED by Windows but fromtimestamp RAISES OSError   <- reachable
```

End to end against the real function, with a `Data.p4k` dated year 4000 (clock skew during a download, an archive restored from a bad backup, a file written from another OS on a shared volume):

```
ranking alone (guarded):   64060588800.0
_pick_live_sc_install:     RAISES OSError: [Errno 22] Invalid argument
```

That propagates out through `_scan_common_sc_install_locations` -> `get_sc_install_root` -> startup. Two things make it worth closing despite the unlikely trigger: the skewed file *wins* the ranking (it is the "newest"), and the log line only runs when `len(candidates) > 1` — so it fires exclusively on multi-install machines, exactly the population this PR exists to serve. Someone in that state would get a crash instead of the diagnostic that was the point.

A small helper next to the existing one keeps the two consistent:

```python
def _p4k_date(mtime: float) -> str:
    if not mtime:
        return "none"
    try:
        return datetime.datetime.fromtimestamp(mtime).strftime("%Y-%m-%d")
    except (OSError, OverflowError, ValueError):
        return "unreadable"
```

### Minor 2 — new flake8 E302

`class AppSettings:` now has one blank line above it rather than two. Confirmed new by running flake8 against the base:

```
base:  1775:5: E303 too many blank lines (5)          <- pre-existing only
this:  220:1:  E302 expected 2 blank lines, found 1   <- introduced here
       1849:5: E303 too many blank lines (5)          <- same pre-existing one
```

`flake8 src/ tests/ scripts/` is a documented quality command; there is no CI gate, which is why nothing caught it.

### Minor 3 — a comment names a function that does not exist

`workers.py:542` says `_raise_unforge_timeout`; the function is `_unforge_timeout_error` (`pak_extractor.py:105`). The return-rather-than-raise rename was deliberate and well argued in the docstring — the comment just didn't follow it. Grep-for-the-name is how people navigate this file.

---

### On the self-reported Major

Your finding — that the detection change can't help a profile that already saved the phantom path, since `get_sc_install_root` returns a valid saved `SC_INSTALL_ROOT` before ever scanning — is correct, and fixing the PR body rather than the code was the right response. I agree with rejecting per-launch re-validation: walking every drive letter at startup for every user to catch a rare case is the wrong trade, and a saved path is often deliberate.

Worth carrying into the **release notes** as well as the PR: 2.3.1 does not repair an already-mispointed profile. What serves those users is the new timeout message, which is why leading with the install path ahead of Verify Files is right — that is the step that actually resolved #370, and Verify Files is what wasted the reporter's time.

---

### Worth calling out

The root-cause writeup is exemplary. Drive-major iteration over a subpath list whose first entry is `Program Files\...` is a genuinely non-obvious interaction, and the table showing which subpath each install matched makes it legible immediately. "The patterns were right. The selection strategy was not" is the correct diagnosis.

The measurements that killed two plausible theories are worth as much as the fix: 175 bytes of stdout rules out `capture_output` buffering, 785 MB peak RSS rules out unforge being a memory hog, and two healthy installs differing by 3.77 GB while producing byte-identical databases rules out archive size as a signal. Recording the theories that died is more useful than recording the one that survived.

One filesystem walk feeding both the ranking and the log — with the comment explaining that a second walk could make the ranked install and the reported date disagree — is the subtle version of that bug rather than the obvious one. `_unforge_timeout_error` returning rather than raising, so `result` isn't statically possibly-unbound, is the kind of reasoning usually left implicit. Recovering unforge's partial output closes a real gap: `subprocess.run` re-collects it onto the exception precisely so a timeout isn't silent, and nothing was catching it.

`DataForgeTimeoutError` follows the `P4kLockedError` precedent exactly, including the reason. 16 tests across both halves cover ties, across-channels, unreadable roots, the install-path-before-Verify-Files ordering, bytes-vs-str decoding, and WARNING-not-ERROR level, both files carrying `unit` + `regression` markers. And translating the message rather than falling back to English is right — a user who can't read the advice can't act on it.

---

### Action items

- [ ] Guard the `fromtimestamp` call in the log line
- [ ] Restore the second blank line before `class AppSettings:`
- [ ] Fix the `_raise_unforge_timeout` name in `workers.py:542`
- [ ] *(release notes)* State that an already-mispointed profile needs a manual Config-tab correction

### Merge order

No overlap with #355/#358/#365. Merges independently.

PR #375, opened 2026-08-27
Recover blueprint names left behind by another localization editor (#372)

### LINKED ISSUE #372: Blueprint tracking
The blueprints for Defiant, Colossus, Endurance, Huracan, Sedulity, Agni, aren't marked as owned under the courier missions "Entry LVL. courier needed in Stanton.

They are in my owned section in the app and others show as owned.

Stealrull:
Fixes #372.

## Summary
- Reported: six blueprints (Defiant, Colossus, Endurance, Huracan, Sedulity, Agni) never showed as owned despite being received, while other items owned around the same time did.
- Root cause: the reporter had previously run StarStrings, another localization editor. Star Citizen's Game.log records whatever name was DISPLAYING at the time, so their log permanently carries StarStrings' naming (e.g. `Ind/1/B Colossus`) instead of ours (`Colossus`). Smart Citizen's log scan imports that verbatim, and it never matches our own item list -- reinstalling doesn't help, since the bad names live in the logs, not in anything we write.
- Fix: `resolve_against_catalogue` (`owned_items.py`) recovers the real item when a scanned name ends in a known catalogue name on a word boundary, whatever decoration precedes it. Deliberately not a StarStrings-specific pattern match -- matching `Ind/1/B` would only fix this one tool and need extending for every editor anyone has used. Wired into the log-scan path, plus a one-shot `_repair_foreign_owned_names` migration that cleans already-broken owned-set entries in place on next launch, since a re-scan alone just re-reads the same old logs.

## Second commit: self-review fixes

A review pass on the first commit found a real defect in it, plus four smaller items. All are fixed in `627953f`.

**The significant one -- the repair step could silently delete a real owned item.** Both recovery paths anchored on `_bp_item_names`: names currently offered as a reward by some loaded mission. That is narrower than "every real item this install knows about", and any item CIG has rotated out of every mission's current reward pool (a normal, recurring state for a long-owned item) would read as unmatched against it. `resolve_against_catalogue` was then free to fold that real-but-unlisted name into an unrelated shorter owned item, and the repair step dropped it as an "already-owned duplicate" -- permanently losing an ownership record. Reproduced directly: a player owning both `Cascade` and `Fierell Cascade`, with only `Cascade` in a partial catalogue, silently lost `Fierell Cascade`. That is the same class of silent loss #372 was filed over, reintroduced by its own fix.

Fixed with `blueprint_meta.known_item_names()` -- every `item_Name`/`vehicle_Name` entry currently loaded, regardless of reward status -- wired into both recovery paths in place of `_bp_item_names`.

Also in that commit:
- Removed `resolve_against_catalogue`'s same-length-tie branch: it is structurally unreachable (a fixed-length suffix of the scanned name has exactly one value, so two distinct catalogue entries can never both match at that length). Docstring corrected to match.
- `_FOREIGN_TAG_BOUNDARY` gains `.` and `:`, so a tool using either as its tag/name divider still recovers.
- `match_import_names` gains an optional `catalogue=`, so Import Owned Blueprints recovers a foreign-tagged name too -- previously only the log scan and the one-shot repair did, meaning an owned set exported before upgrading round-tripped as permanently unmatched.
- `_repair_foreign_owned_names` is now wrapped defensively: it runs unconditionally from `_rebuild_blueprint_metadata`, so a failure in this ancillary cleanup must never surface as a load/merge failure after the entries have already loaded.
- Extracted the decision logic into a pure `owned_items.repair_foreign_owned_names()`, so the MainWindow entry point that mutates persisted settings (previously untested) is directly covered.

## Tests
- `tests/test_foreign_editor_blueprint_names.py` -- 26 tests: the six reported names, four other class prefixes seen in the wild (Civ/Mil/Sth/Cmp), four formats no tool uses yet, the mid-word guard, longest-match, idempotency, the new boundary characters, the partial-catalogue safety cases, and an end-to-end pass against `tests/fixtures/kraken_global_latest.ini` (a StarStrings-modified fixture with 451 real tagged names) -- zero wrong recoveries.
- `tests/test_repair_foreign_owned_names_wiring.py` -- 6 tests for the MainWindow entry point: that it reads the wide catalogue rather than the narrow one, that a partial catalogue leaves real owned items alone, that a clean set writes nothing, and that a failure inside the repair neither propagates nor half-writes.

Full suite on this branch: **1649 passed, 21 skipped, 0 failed.**

## Note on merge order
This PR conflicts with #355 and #358 (all three orderings; see the comment thread). Resolution is keep-both in every case. If this lands after #355, `known_item_names()` should also take `enclosings`/`default_values` and forward them to `normalize_item_name`, matching `build_blueprint_metadata` -- that parameter does not exist on this branch yet. Already proven on the local integration branch.


Osiris-DevWorks (dismissed):
## Review — #372 foreign-editor blueprint names

Approving. **1649 passed, 21 skipped, 0 failed.** The design is right, and the self-review caught a genuine data-loss defect in the first commit. One residual case in the same code path is worth closing before merge.

**The approach is the correct one.** Anchoring on "does this end in a known real item name, on a word boundary" rather than pattern-matching StarStrings' `Ind/1/B` is what makes this work for the next tool nobody has heard of yet. The longest-match rule and the reasoning for why two catalogue entries can never tie at the same length (a fixed-length trailing slice has exactly one value, so at most one set member can equal it) are both correct — and removing the unreachable tie branch rather than leaving it as defensive noise was the right call.

Verified the recovery works: `resolve_against_catalogue("Ind/1/B Colossus", {"Colossus"})` -> `Colossus`.

### Minor — the drop path can still delete a real owned item

`repair_foreign_owned_names` discards an entry outright when the resolved name is already owned:

```python
repaired.discard(nm)
if real in repaired:
    renamed[nm] = None      # dropped, no undo
```

Switching to `known_item_names()` fixed the reward-rotation trigger, and `test_partial_catalogue_does_not_delete_a_real_owned_item` locks that — but it builds the catalogue with *both* items present, so it validates the fixed path rather than the residual one. The algorithm still deletes whenever the catalogue is incomplete for any *other* reason.

I tested the two triggers I could think of:

```
translated run (Italian catalogue, English owned):  repaired=['Cascade','Fierell Cascade']  no loss
CIG removes/renames the longer item's loc key:      repaired=['Cascade']  renamed={'Fierell Cascade': None}  LOST
```

The language case is safe — good. The second is not: an owned item whose loc key disappears from the game data, whose name is a boundary-suffix of another owned item, is silently and permanently dropped. Narrow, but CIG does remove and rename loc keys between patches, and this runs on every load.

It also sits awkwardly against this function's own stated principle, which I think is the right one:

> *"An unrecoverable name is left exactly as-is rather than dropped: it may be a real item this install simply cannot see right now, and silently deleting a user's owned marks is worse than leaving one unmatched."*

That reasoning applies just as well to a name that *was* resolved but collides with an existing entry. The cheapest fix is to stop dropping — keep the original alongside, and let a genuine duplicate show as one harmless extra entry rather than risking a real one:

```python
if real in repaired:
    repaired.add(nm)        # keep both; a stray duplicate beats a silent deletion
    continue
```

If you'd rather keep the dedupe, logging that branch at WARNING instead of INFO would at least leave a trail.

### Also good

`known_item_names` being separate from `build_blueprint_metadata`'s keys, with the docstring spelling out that "absent from the catalogue" must mean "not a known real item" and never "known but temporarily unlisted", is the distinction the whole fix rests on. Wrapping `_repair_foreign_owned_names` defensively is right — the entries have already loaded by then, so an ancillary cleanup must never surface as a load failure. Extending `match_import_names` with `catalogue=` closes the round-trip gap where an owned set exported before upgrading came back permanently unmatched. And the end-to-end pass over the StarStrings-modified fixture with 451 real tagged names and zero wrong recoveries is the evidence that matters most.

### Merge order

Your note is right that this conflicts with #355 and #358 three ways. Since #355 needs another commit anyway (its `parse_component_tag` gate), the ordering that costs least is **#358 -> #355 -> this**, with `known_item_names()` picking up `enclosings`/`default_values` on the way past.

PR #376, opened 2026-08-27
Apply search/dropdown filters to the Owned list too (#374)

### LINKED ISSUE #374: Blueprint tracker dropdowns don't apply to owned items
Dropdowns or searching with specific names should filter owned items as well as not owned, currently it doesn't.

Stealrull:
Fixes #374.

## Summary
- Reported: the Blueprint Tracker's search box and Type/Class/Size/Grade/Mission dropdowns only ever filtered the Available list -- narrowing down had no effect on the Owned list at all.
- Root cause: `_refilter_blueprints_available` (wired to every filter control's signal) only ever iterated `self._blueprints_available_list` and hid rows there. The Owned list was rendered once from the full owned set and never touched again.
- Fix: `_blueprint_item_visible(name)` is already a pure predicate with nothing list-specific about it, so renamed the method to `_refilter_blueprint_lists` and applied it uniformly to both `QListWidget`s.

## Tests
`src/gui/blueprint_tracker_tab.py` is GUI wiring (widget layout, signal connections), exempt from unit tests per this repo's own coverage policy -- manual smoke test is the documented path. Confirmed working in a portable build: search and each dropdown now narrow both lists.

Full suite: 1614 passed, 21 skipped, 0 failed on this branch.


Stealrull (commented):

Small, contained diff -- one method renamed and its body widened from one list to two.

**Correctness:** `_blueprint_item_visible(name)` reads only the search text, the mission combo, the facet combos and `self._blueprint_meta` -- nothing about which list a row lives in. Applying it to both `QListWidget`s is therefore the same predicate over a second collection, not a behaviour change to the predicate itself.

**Call sites:** all four references updated (search `textChanged`, mission combo, the 4 facet combos in the loop, and the `_render_blueprint_lists` call after a repopulate), so no stale name is left pointing at the old method.

**Considered and deliberately unchanged:** a hidden-but-selected Owned row stays selected, so the arrow button would still act on it. That is pre-existing, symmetric behaviour -- the Available list has worked exactly this way since filters were added -- so aligning the two lists doesn't introduce it. Worth a separate look if it ever bites someone, not something to fold into this fix.

**Test coverage:** `src/gui/blueprint_tracker_tab.py` is GUI wiring, exempt per `.claude/commands/test_coverage_check.md` (manual smoke test is the documented path). Confirmed working in a portable build -- search and each dropdown now narrow both lists. Full suite green.

No blocking issues.

Osiris-DevWorks (changes requested):
## Review — #374 filter the Owned list too

The change itself is exactly right: `_blueprint_item_visible` is a pure predicate on the name, so applying it to both lists is the correct generalisation, the rename is complete (no stale references to `_refilter_blueprints_available` anywhere in `src/` or `tests/`), and the docstring explains the user-facing problem rather than restating the code.

Your test-exemption call is also correct — I checked `.claude/commands/test_coverage_check.md`, and `src/gui/*.py` (widget layout, signal connections) is explicitly **Exempt**, "Surface for review". So no coverage finding here.

But filtering the Owned list exposes a hazard that could not happen while that list never filtered.

### Major — a stale selection under a filter lets Remove un-own invisible items

Hiding a `QListWidgetItem` does not deselect it, and both lists are `ExtendedSelection` (`blueprint_tracker_tab.py:321`, `:349`). `_selected_names` reads `selectedItems()` with no visibility check, so:

```
user selects 3 items      : ['Colossus', 'Defiant', 'Endurance']
after filter hides 2      : visible=['Colossus']
what 'Remove' would unown : ['Colossus', 'Defiant', 'Endurance']
```

Reproduced against the real selection mode. The sequence is an ordinary one — select a few, realise you want to narrow down, type in the search box, click Remove — and the two items the user can no longer see lose their ownership record silently.

This shape existed for the Available list before this PR, but there it only ever *added* ownership, which is recoverable and visible. Extending the filter to the Owned list turns it into deletion, which is the same class of silent ownership loss #372 was filed over.

One line, and it fixes both lists at once:

```python
def _selected_names(self, lst) -> list:
    return [it.data(Qt.ItemDataRole.UserRole)
            for it in lst.selectedItems() if not it.isHidden()]
```

Clearing the selection inside `_refilter_blueprint_lists` would also work, but it throws away a selection the user may want back when they clear the filter — filtering at the point of use is the smaller change and keeps "what you can see is what you act on" true by construction.

### Minor — no empty-state feedback when a filter empties both lists

`_render_blueprint_lists`'s empty-state branch keys on there being no metadata and nothing owned, so a filter that matches nothing now leaves two blank lists with the controls still shown and no indication why. Not new behaviour for Available, but it is newly reachable for Owned, and two empty panes read more like a bug than a filter result. Worth a "no matches" hint if it is cheap; not worth holding the PR for.

Requesting changes on the selection item only — the rest of the change is good and I would take it as-is otherwise.

Osiris-DevWorks (approved):
## Re-review — fix confirmed, lifting the block

`Exclude hidden rows from selection (#374 review)` is exactly the right shape, and filtering at the point of use rather than clearing the selection on every filter change is the better of the two options — it keeps a selection the user may want back once they clear the filter.

Verified against the original reproduction rather than by reading the diff:

```
user selects 3            : ['Colossus', 'Defiant', 'Endurance']
filter leaves visible     : ['Colossus']
what Remove would un-own  : ['Colossus']
-> hidden items protected : True
```

Before the fix that last line read `['Colossus', 'Defiant', 'Endurance']`, silently un-owning two items the user could no longer see.

The docstring naming this as the same class of silent ownership loss as #372 is a good call — it is, and recording that connection is what stops someone "simplifying" the visibility check away later.

Branch merged up to the integrated 2.3.1 tip (#358, #355, #375, #377) so CI ran against what it actually merges into. Full suite **1784 passed, 21 skipped, 0 failed**.

The empty-state Minor from my first pass (two blank lists when a filter matches nothing, with no "no matches" hint) is unaddressed and still non-blocking — worth a look sometime, not worth holding this.

PR #377, opened 2026-08-27
Rewrite the tester Test Plan for the 2.3.1 scope

Stealrull:
## Summary
- The in-app Test Plan still carried 2.3.0's checklist: new languages, Settings Backup, ship favoriting, the RS ore-name work. A tester on a 2.3.1 pre-release build was being pointed at last release's surface and at nothing in this one.
- `src/utils/CLAUDE.md` calls for `TEST_SECTIONS` to track the active release's diff over its base, updated per release, so this was a stale-content gap rather than a code bug.

## What's in it now
One section per 2.3.1 area:
- **Tag Builder enclosing styles (#352)** - all five styles including None (space only), plus a check that Square still behaves as it did in 2.3.0.
- **Renamed blueprints mission header (#353)** - the tracker, the BP Descriptions filter, and the `[Owned]` weave.
- **Blueprint names from another editor (#372)** - the one-shot repair on launch, and the "nothing you owned disappears" guarantee.
- **Blueprint Tracker lists and filters (#354, #374)**.
- **Window layout and String Editor columns (#364)** - the largest section.
- **Star Citizen install detection (#370)**.
- **Translated runs and channel switching (#363, pool headings)**.

Core workflow (smoke) and the portable-build checks are kept as-is: both are evergreen rather than release-scoped.

`plan_hash()` changes with the content, so testers' stale 2.3.0 check-marks are dropped rather than silently re-pointed at different items.

## Note for the next person editing this
Every item is under 190 characters. `tests/test_test_plan.py` chunks the submitted report at a 200-char limit and asserts each line survives intact, so a longer item fails that test rather than just wrapping awkwardly. That constraint isn't obvious from the code, so it's now a comment above `TEST_SECTIONS`.

## Tests
Full suite: 1765 passed, 21 skipped, 0 failed (on the integration branch with every pending 2.3.1 PR merged). Verified in a portable build.

### DIFF
--- a/src/utils/test_plan.py
+++ b/src/utils/test_plan.py
@@ -19,7 +19,11 @@ import json
 
 # Each section is a title plus a flat list of one-line test items. Keep items
 # imperative and self-contained ("do X, confirm Y") so a tester needs no other
-# doc. This plan covers Smart Citizen 2.3.0 (the diff over its 2.2.1 base).
+# doc. This plan covers Smart Citizen 2.3.1 (the diff over its 2.3.0 base).
+#
+# Keep each item under ~190 characters. tests/test_test_plan.py chunks the
+# submitted report at a 200-char limit and asserts every line survives intact,
+# so a longer item fails that test rather than just wrapping awkwardly.
 TEST_SECTIONS: list[dict] = [
     {
         "title": "Core workflow (smoke)",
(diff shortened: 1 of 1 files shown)


Osiris-DevWorks (approved):
## Review — 2.3.1 Test Plan rewrite

Approving. Content-only change to `TEST_SECTIONS`, which `src/utils/CLAUDE.md` explicitly calls for per release, so this is the documented maintenance rather than a code change.

**Verified on the branch:**

```
sections=9  items=46  longest=182 chars  items over 190: 0
tests/test_test_plan.py: 12 passed
```

So the 190-character claim holds with headroom, and the comment recording *why* that limit exists — `tests/test_test_plan.py` chunks at 200 and asserts each line survives intact — is the useful part. That constraint is genuinely invisible from the code, and the failure mode (a test failure rather than an ugly wrap) is not what anyone would guess.

**Scope coverage checks out** against the actual 2.3.1 work: #352, #353, #372, #354/#374, #364, #370, #363 and the pool headings each get a section, and keeping Core workflow and the portable-build checks as evergreen is right — they are not release-scoped and testers need them every time.

The items are written the way a test plan should be: concrete objects a tester can actually look for (`NDB-26 Repeater`, `QuadraCell MT`, `FR-66`) rather than "verify tags work". The #372 section leading with "check the Owned list for garbled names like `Ind/1/B Colossus`" and closing on "nothing you owned should disappear" is particularly good — it gives the tester both the symptom and the safety property, and the second is the one that actually needs eyes.

Letting `plan_hash()` drop testers' 2.3.0 marks rather than silently re-pointing them at different items is the correct behaviour and worth having said out loud.

### One sequencing note, not a change request

Every section here except Core workflow and portable describes behaviour from PRs that are still open (#355, #358, #365, #371, #375, #376). If this merges into `release/2.3.1` ahead of them and someone adds `build-installer` to a PR in the meantime, the tester gets a preview build whose in-app Test Plan describes work that is not in it — items that cannot pass, with no way to tell that from a real regression.

Worth **merging this last**, after the PRs it documents. Or if it lands early, it is worth a word to testers that the plan is ahead of the build.

PR #384, opened 2026-08-29
Add opt-in "delete all saved settings" checkbox to the uninstaller (#357)

### LINKED ISSUE #357: Add a delete all saved settings tickbox to uninstaller

Stealrull:
Fixes #357.

Replaces #359, opened from the fork before the branch-off-upstream policy took effect on 2026-08-27. Same code, this branch just lives on the main repo instead so `claude-review` can actually run on it. #359's review discussion stays there for reference.

## Summary
- Uninstalling has always deliberately left the registry node, user data folder, and DataForge cache behind (the "#172 persistence lock") so a reinstall picks up where the user left off. That's the right default, but leaves no way to get a genuine clean slate when something has gone wrong -- a Discord user reported exactly this friction.
- Adds a custom uninstall confirmation dialog with an unchecked-by-default "Also delete all saved settings" checkbox. Ticking it pops up a separate warning listing exactly what gets deleted, requiring explicit OK to proceed -- declining reverts the checkbox rather than silently continuing checked.
- Only when explicitly confirmed does the uninstaller delete all three locations named in the issue:
  1. `HKEY_CURRENT_USER\Software\Osiris DevWorks\Smart Citizen`
  2. The user data folder -- override-aware (`user_data_dir`), not just the `Documents\Smart Citizen` default
  3. The separate DataForge cache folder -- also override-aware (`cache_dir`); this one was never cleaned by the normal uninstall path at all, custom location or not, so this closes a second, smaller pre-existing gap too
- Leaving the box unchecked is byte-for-byte the existing uninstall behavior. The `#172` persistence-lock comment block is preserved and extended to document this as its one deliberate, explicitly-gated exception.
- The dialog is skipped entirely during a silent uninstall (`UninstallSilent()`) -- the in-app auto-updater's "remove old version" step runs its uninstaller with `/VERYSILENT`, and a raw custom VCL form isn't auto-suppressed by Inno's silent flags the way `MsgBox` is. Without this guard an automated upgrade would hang forever waiting for a click nobody will make.
- Follow-up commit addresses #359's own review feedback: the cache-folder skip check now reuses `PathUnderRoot` (already used for OneDrive detection in this file) so it catches `cache_dir` nested under the user data folder, not just an exact match. No behavioural change -- the nested case already fell through to a harmless no-op -- just removes a redundant delete attempt and its confusing log line.

## Test plan
- [x] `ISCC.exe installer.iss` compiles clean (no Pascal Script errors) -- this feature has no Python code, so the compiler is the only automated check available.
- [x] Manual test, two rounds: first round found the checkbox caption running off the dialog (`TNewCheckBox` doesn't word-wrap like a text label) and asked for the warning to be conditional/require explicit acknowledgment rather than static always-visible text. Second round (after moving the warning into a popup fired on check, with revert-on-cancel) confirmed working as intended.
- [x] Rebased onto current `main` (2.3.1) and recompiled clean on the new base.

### DIFF
--- a/installer.iss
+++ b/installer.iss
@@ -108,6 +108,29 @@ begin
   Result := ExpandConstant('{param:AUTOUPDATE|0}') = '1';
 end;
 
+function IsVerySilentUninstall(): Boolean;
+var
+  I: Integer;
+begin
+  { #357 review fix: UninstallSilent() is True for BOTH /SILENT and
+    /VERYSILENT, but only /VERYSILENT means genuinely headless -- /SILENT
+    still shows MsgBox prompts, and the "click NO, uninstall old version
+    only" branch of the reinstall-detection prompt below runs its
+    sub-uninstall with exactly /SILENT (an interactive user is sitting
+    right there, mid-click). Checking the actual command line instead of
+    UninstallSilent() distinguishes that case from the truly automated
+    /VERYSILENT auto-update cleanup (UnInstallOldVersion above). }
+  Result := False;
+  for I := 1 to ParamCount do
+  begin
+    if CompareText(ParamStr(I), '/VERYSILENT') = 0 then
+    begin
+      Result := True;
+      Exit;
+    end;
+  end;
+end;
+
 function GetLocalCacheDefault(): String;
 begin
   { Mirrors AppSettings.get_dataforge_cache_base() default for registry
@@ -854,7 +877,7 @@ end;
 
 procedure DeleteAllUserSettings();
 var
-  UserDataDir, CacheDir: String;
+  UserDataDir, CacheDir, LegacyUserDataDir: String;
 begin
   { #357: the ONE deliberate exception to the #172 persistence lock
     documented in CurUninstallStepChanged below — only reached when the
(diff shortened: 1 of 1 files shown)


Stealrull (commented):
## Review verdict: Approve

**What:** Adds an opt-in "Also delete all saved settings" checkbox to the uninstaller, closing #357. Unchecked (default) is byte-identical to existing behavior. Checked, with an explicit acknowledge-to-proceed popup, deletes the HKCU registry node, the (override-aware) user data folder, and the (override-aware, previously never-cleaned) DataForge cache folder.

**Issue acceptance check (#357):** all criteria satisfied — opt-in checkbox default-off, warning required before it takes effect, all three named locations (registry node, user data folder incl. overrides, DataForge cache folder incl. overrides) deleted only on explicit opt-in.

**File review — `installer.iss`:** Clean.
- Traced `GetCacheDir()`'s registry key (`cache_dir`) and `GetLocalCacheDefault()`'s path against `src/utils/settings.py`'s `CACHE_DIR` constant and `get_dataforge_cache_base()` — both match exactly.
- Confirmed path-override resolution happens before the registry node (which holds those overrides) is deleted — correct ordering.
- Verified the two post-review fixes against the file's own two real invocation sites: `IsVerySilentUninstall()` correctly returns True for the genuinely headless `/VERYSILENT` auto-update cleanup and False for the interactive "click NO, uninstall old version only" `/SILENT` reinstall flow (so the dialog now correctly shows there instead of being wrongly suppressed). The legacy `Documents\SC Localization Editor\` folder cleanup mirrors the existing `MigrateUserDocsFolder` pattern exactly.
- No DRY, threading (N/A — Pascal Script), or docs-drift issues found (checked `docs/HELP.md`/`docs/ABOUT.md`, no uninstall mentions to go stale).

**Test coverage:** No Python test infra applies to Pascal Script (same exemption class as `scripts/`/GUI-wiring). Verified manually this session: `ISCC.exe installer.iss` compiles clean, both invocation sites hand-traced against the new logic, and full `pytest tests/` run as a sanity check on the rest of the tree — 1786 passed, 21 skipped, no failures.

No Critical/Major/Minor issues found.

PR #391, opened 2026-09-03
Fix flaky test suite: share one MainWindow per file, not per test

### LINKED ISSUE #390: Flaky test suite: test_tab_scrollbar_placement.py fails intermittently depending on run order
## What's happening

`tests/test_tab_scrollbar_placement.py`'s 4 tests intermittently fail when run as part of the full suite (`pytest tests/ -q`), but always pass in isolation (`pytest tests/test_tab_scrollbar_placement.py -q`). The failure is a Qt window-flags bitmask assertion mismatch.

Reproduced this on multiple unrelated branches this cycle, including one that never touches any GUI/tab/scrollbar code at all -- confirms it's test-order pollution, not a real bug in the scrollbar placement fix itself.

## Root cause

`tests/test_branding_labels_layout.py`'s `window` fixture was function-scoped, so its 3 tests each built a full real `MainWindow` via `build_main_window`. `tests/gui_window.py` never destroys these (documented as unsafe -- `close()` + `deleteLater()` causes access violations), so all 3 windows stayed alive in `_LIVE_WINDOWS` for the rest of the process.

Since `test_branding_labels_layout.py` sorts alphabetically before `test_tab_scrollbar_placement.py`, by the time the scrollbar tests build their own window there were 4 concurrent real top-level `MainWindow`s alive -- the same exposure level `gui_window.py`'s own docstring blames for the original intermittent crashes ("measured 1 run in 6 with four constructions").

`test_tab_scrollbar_placement.py` was deliberately made module-scoped (1 construction) to avoid this, but `test_branding_labels_layout.py` never got the same treatment -- its docstring's claim "This is the only test that builds a real MainWindow" had gone stale once the scrollbar test file was added.

## Fix

Made `test_branding_labels_layout.py`'s `window` fixture module-scoped (one construction shared across its 3 tests), mirroring `test_tab_scrollbar_placement.py`'s existing pattern (`tmp_path_factory` + `MonkeyPatch.context()`). Updated the stale docstrings in both files to reflect that two files now build real windows, not one.

Verified clean across several full-suite runs after the fix (1915 passed, 21 skipped each time). Couldn't get a red repro even before the fix on the machine that found this, so it's a structural fix matching the codebase's own documented failure mode rather than a red-to-green repro -- worth a few extra CI runs before calling it fully closed.

Not a functional bug in the app itself, just test-suite reliability.

Stealrull:
## Summary

- `tests/test_branding_labels_layout.py`'s `window` fixture was function-scoped, so its 3 tests each built a full real `MainWindow`. `tests/gui_window.py` never destroys these (documented as unsafe -- `close()` + `deleteLater()` causes access violations), so all 3 stayed alive in `_LIVE_WINDOWS` for the rest of the process.
- That file sorts alphabetically before `test_tab_scrollbar_placement.py`, so by the time the scrollbar tests built their own window there were 4 concurrent real `MainWindow`s alive -- the same exposure level `gui_window.py`'s own docstring blames for the original intermittent crashes ("measured 1 run in 6 with four constructions").
- `test_tab_scrollbar_placement.py` was already made module-scoped to avoid this; `test_branding_labels_layout.py` never got the same treatment, so its docstring's claim ("This is the only test that builds a real MainWindow") had gone stale.
- Fixed by making the `window` fixture module-scoped, one construction shared across the file's 3 tests, mirroring `test_tab_scrollbar_placement.py`'s existing pattern (`tmp_path_factory` + `MonkeyPatch.context()`).
- Review follow-up: the first pass left `show()`/`resize()` calls inside two of the three test bodies rather than the fixture, so the second test's resize became a same-size no-op silently reusing state the first test already established -- an accidental order-coupling between tests meant to be independent. Moved the one-time `show()`/`resize()` into the fixture itself (matching `test_tab_scrollbar_placement.py` exactly); both tests now pass independently in isolation as well as together.

Not a functional bug in the app -- purely test-suite reliability. No `src/` changes.

Fixes #390

## Test plan

- [x] Full suite (`pytest tests/ -q`) passes: 1786 passed, 21 skipped
- [x] Stable across several repeated full-suite runs after the fix
- [x] Both previously-coupled tests (`test_branding_labels_do_not_grow_with_the_window`, `test_tagline_stays_directly_under_the_title`) pass independently in isolation, not just together
- [x] Mutation-tested against the original bug: reverted the `Fixed` size policy back to `Preferred` in `main_window.py`, confirmed `test_branding_labels_have_a_fixed_height_policy` still catches it, then restored the fix

### DIFF
--- a/tests/test_branding_labels_layout.py
+++ b/tests/test_branding_labels_layout.py
@@ -12,12 +12,18 @@ The fix is a Fixed vertical policy on both. It is one declarative line in
 setup_ui, so without this test a regression that dropped it would reinstate
 the reported bug with the whole rest of the suite still green.
 
-This is the only test that builds a real MainWindow. It is worth the cost
-here because the bug is a property of the assembled window: the stub used
-elsewhere in tests/test_ui_mode.py has no branding labels and no content
-layout to distribute height through, so it cannot observe this at all.
-Settings, user-data and cache directories are all redirected to tmp_path so
-the construction touches nothing real.
+This is one of only two test files that build a real MainWindow (see also
+tests/test_tab_scrollbar_placement.py). It is worth the cost here because the
+bug is a property of the assembled window: the stub used elsewhere in
+tests/test_ui_mode.py has no branding labels and no content layout to
+distribute height through, so it cannot observe this at all. Settings,
+user-data and cache directories are all redirected to tmp_path so the
+construction touches nothing real.
+
+The ``window`` fixture below is module-scoped for the same reason
+tests/test_tab_scrollbar_placement.py's is: see tests/gui_window.py for why
+building one per test destabilised the whole run. Keep it that way rather
+than reverting to a function-scoped fixture per test.
 """
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


Stealrull (commented):
## Review verdict: Approve

Reviewed via /review_pr — file-by-file walkthrough, checkpointed. (Posted as a comment, not a formal GitHub approval -- GitHub blocks self-approval since this is also the PR author's account.)

**Files reviewed (1):**
- `tests/test_branding_labels_layout.py` — Clean. Module-scoped `window` fixture (one `MainWindow` shared across the file's 3 tests instead of one per test), matching `test_tab_scrollbar_placement.py`'s existing pattern. The one-time `show()`/`resize()` settle moved into the fixture itself, so the two tests that used to redundantly call them are no longer implicitly order-coupled -- verified passing independently in isolation as well as together.

**Standards check:** no Smart Citizen-specific conventions apply (no QSettings, table columns, worker threads, DataForge subtrees, version stamp, or base.ini/global.ini naming touched).

**Issue acceptance (#390):** root cause fixed, verified stable across repeated full-suite runs, mutation-tested against the original bug. CI green on this PR.

No issues found -- Critical: 0, Major: 0, Minor: 0.

PR #392, opened 2026-09-03
Materialise setting defaults into Export Settings backups (#383)

### LINKED ISSUE #383: Export settings doesn't save default settings
Export settings currently doesn't take into account default settings so if a user exports then changes something and wants to import their old settings it won't replace things that where changed after the export if they where changed from a default.

Need to expand the export save file to include defaults as well as changes so the user can truly restore their setup, default or not.

Stealrull:
Fixes #383.

## Summary
- Settings backups only ever captured keys that had been explicitly changed from default — `QSettings`/`JsonSettings` never store an unset value, so anything still at its default had no key for `export_all_values()` to enumerate.
- That made an import unable to restore a preference *back* to its default: rename a Mission Label, import an older backup, and the rename survived instead of reverting. Same gap existed for theme, UI mode, the favourites prefix, and every checkbox.
- `AppSettings.profile_default_values()` now lays down every preference's true default before `export_all_values()` overlays the real stored values on top, so a backup describes the user's whole configuration, not just what differs from stock. Stored values still win over defaults.
- `PROFILE_DEFAULT_EXCLUDE_KEYS` / `PROFILE_DEFAULT_EXCLUDE_PREFIXES` carve out per-machine state and user *data* that must stay excluded so an import restores rather than destroys: install paths (clearing one forces re-detection, which can land on the wrong install per #370), `owned_items` (the player's actual blueprint collection), the blueprint log watermark, and the test-plan keys.
- `_MISSION_HEADER_SETTING` extracts the field→key mapping that three places needed (`get_mission_headers`, `set_mission_header`, `profile_default_values`) — no behavior change, just removes the duplicated dict.
- `tests/test_settings_profile_defaults.py::TestNoDrift` replays every no-arg `AppSettings.get_*` getter against a recording backend and fails if a defaulted key is neither materialised nor deliberately excluded, so a newly added setting can't silently reintroduce this bug.

## Follow-up (review pass)
- `is_profile_default_materialised(key)` called `is_profile_excluded_key(key)` with no value, so its `data_sources/*/path` URL check couldn't tell a URL-mapped default from a local one and fell back to treating any such key as non-URL (excluded). The value was already available at the one call site in `export_all_values()` — just wasn't threaded through. No default in `profile_default_values()` has that shape today, so this was a no-op in practice; closed the gap anyway since passing it costs nothing.

## Test plan
- [x] New `tests/test_settings_profile_defaults.py` — defaults travel in the export, stored values still override, the issue's own export→change→import scenario now restores correctly, exclude keys/prefixes stay out, and the drift guard passes.
- [x] `pytest tests/` — 1804 passed, 21 skipped, no failures.


Stealrull (commented):
## Review verdict: Approve

**What:** Settings backups only captured keys explicitly changed from default (`QSettings`/`JsonSettings` never store an unset value), so an import couldn't restore a preference *back* to its default — reported as a Mission Label rename surviving import of an older backup. `AppSettings.profile_default_values()` now materialises every preference's true default before `export_all_values()` overlays real stored values on top.

**Issue acceptance check (#383):** both stated criteria satisfied — defaults now travel in the export, and stored (changed) values still take priority over defaults.

**File review:**
- `src/utils/settings.py` — Clean. Traced `profile_default_values()`'s entries against the real getter defaults they mirror (e.g. `MERGE_HIERARCHY` matches `get_merge_hierarchy()`'s actual default exactly). Confirmed `is_profile_default_materialised()`'s one-arg call into `is_profile_excluded_key(key)` is safe (`value` defaults to `None` there, and the branch that uses it never fires against anything this function produces). `_MISSION_HEADER_SETTING` extraction is a legitimate DRY win (3 call sites) and behavior-preserving.
- `tests/test_settings_profile_defaults.py` — Clean. `TestNoDrift::test_every_getter_default_is_accounted_for` is the standout: it replays every real `get_*` getter and fails if a defaulted key isn't materialised or excluded, so a future setting can't silently reintroduce this bug.
- `tests/test_settings_profile.py` — Clean. Three existing assertions correctly loosened from exact-dict/magic-number checks to presence/absence and self-consistent checks, since the new default-carrying export would otherwise make them brittle.
- `src/utils/CLAUDE.md` / `tests/CLAUDE.md` — Clean. Cross-checked every symbol name cited (`_MISSION_FIELD_SETTING`, `_MISSION_TITLE_TAG_SETTING`, `_MISSION_TITLE_TAG_DEFAULTS`, etc.) against the actual code — all exist exactly as named.
- `docs/HELP.md` + 7 translated `HELP.md` files — Clean. New sentence accurately describes the behavior (including the sharp edge: a post-backup change reverts on restore). Structurally verified each translation lands the addition in the same position with consistent terminology and clean UTF-8.

**Test coverage:** Required area (`src/utils/`), well covered. Ran the new test file plus full suite locally before push: 1804 passed, 21 skipped, no failures.

No Critical/Major/Minor issues found.

Issue #385, opened 2026-08-29
Confirm Star Citizen install location


Stealrull:
Extra button under Browse for install directory which checks the CIG launcher for the in use Star Citizen install location and confirms to the user they have smart citizen pointed to the right directory, this will stop issues of users who have phantom installs.

Stealrull:
Fixed in PR #393

PR #394, opened 2026-09-03
Add opt-in auto-scan for owned blueprints on startup (#386)

### LINKED ISSUE #386: Auto-Update Owned Blueprints
Add a process by which every time you start Smart Citizen, it can auto-update from logs your owned blueprints. This is an opt-in checkbox in the tracker.

Stealrull:
Fixes #386.

## Summary
- Adds an opt-in **Auto-scan logs on startup** checkbox to the Blueprint Tracker tab. When enabled, "Scan Logs for Owned Blueprints" runs automatically on every launch instead of only on a manual click, so the owned set stays current without a click. Off by default.
- Reuses the exact same scan pipeline a manual click drives (channel queue, watermark, #268 multi-channel coverage via **Also scan LIVE/HOTFIX** — the auto-scan and the manual button call the identical `_channels_to_scan()` helper with the identical `AppSettings.get_scan_other_channels_enabled()` setting, so that checkbox applies to both the same way) — the only difference is presentation: a silent run reports through the status bar instead of the manual scan's modal progress dialog and summary popup, since a launch isn't the place for a popup interruption.
- Follow-up **Show popup for new blueprints found by auto-scan** checkbox (off by default) opts back into the manual scan's own summary dialog for anyone who wants the interruption anyway — only applies when the auto-scan actually finds something new; a quiet run stays silent either way.
- `BlueprintTrackerTab.set_scan_logs_enabled()` (new) disables the "Scan Logs" button with an explanatory tooltip while any scan is running, manual or silent — closes a gap where a click during an in-flight silent auto-scan hit the pre-existing reentrancy guard and did nothing visible.

## Known follow-up (cross-PR, not fixable here)
This PR adds two new `AppSettings` booleans (`BLUEPRINT_AUTO_SCAN_ON_STARTUP`, `BLUEPRINT_AUTO_SCAN_SHOW_POPUP`). #392 (closing #383) adds `profile_default_values()`, the map of every setting's default for settings-backup exports, with its own `TestNoDrift` test that fails if a defaulted setting isn't accounted for there. Neither PR can register these two keys against the other right now — #392's `profile_default_values()` doesn't exist on this branch's base, and this PR's new constants don't exist on #392's base, so making either PR reference the other would create a dependency between two otherwise-independent branches. Once both are merged into `release/2.4.0`, `TestNoDrift` will fail until a small follow-up adds these two keys to `profile_default_values()` (or explicitly excludes them) — flagging it here so that's an expected, not surprise, next step.

## Test plan
- [x] New `tests/test_blueprint_auto_scan.py` — the two settings' get/set contracts, `_maybe_auto_scan_blueprints`'s gating (off, no install path, already-scanning, valid path), the silent-mode status-bar reporting path (with and without new names), and the popup opt-in's interaction with both the silent and manual paths.
- [x] `pytest tests/` — 1803 passed, 21 skipped, no failures.


Stealrull (commented):
## Review verdict: Approve

**What:** Adds an opt-in "Auto-scan logs on startup" checkbox that runs the existing "Scan Logs for Owned Blueprints" pipeline automatically on every launch, plus a follow-up "Show popup" checkbox to opt back into the manual scan's modal summary. Both off by default.

**Issue acceptance check (#386):** both stated criteria satisfied — automatic per-launch owned-blueprint updates from logs, gated behind an opt-in checkbox in the tracker.

**File review (22 files, all Clean):**
- `settings.py` — two new getter/setter pairs follow the file's established boolean-setting pattern exactly.
- `blueprint_tracker_tab.py` / `main_window.py` — traced the full gating chain (`_maybe_auto_scan_blueprints`'s reentrancy guard, install-path check, silent-flag lifecycle through `_finish_blueprint_scan_queue`). Confirmed the auto-scan reuses the *identical* `_channels_to_scan()` call and `AppSettings.get_scan_other_channels_enabled()` setting as the manual scan — not a parallel implementation, so "Also scan LIVE/HOTFIX" and the watermark apply the same way to both. Confirmed the error path (`BlueprintLogScanWorker` always emits `finished(None)` after `error`) can't strand the button-disabled/silent-flag state. This feature already applies the exact re-entry-guard lesson found the hard way in #385 — the "Scan Logs" button visibly disables with an explanatory tooltip during any in-flight scan, silent or manual.
- Tests (`test_blueprint_auto_scan.py`, 17 cases) — genuinely targeted, covering every branch traced by hand during review.
- `tests/CLAUDE.md` / `src/gui/CLAUDE.md` — accurate, correctly added from the start (unlike #385, no gap here).
- Docs + all 7 translated `HELP.md`/`ui.json` — found and fixed one real gap during this review: the PR initially updated only the English `docs/HELP.md`, leaving all 7 translated files silently missing the new feature's section (not a whole-file fallback, since the files exist — a non-English reader would never see it described). Fixed with translations matching each language's own established terminology, using the exact checkbox labels already in each `ui.json`. Post-fix, zero i18n key drift across all 8 languages.

**Known, self-catching follow-up:** two new settings aren't yet registered in #392's `profile_default_values()` — can't be fixed in either PR alone without creating a false dependency between two independent branches (each references constants that don't exist on the other's base). Flagged in this PR's own description; #392's own `TestNoDrift` guard will catch it automatically once both merge.

**Test coverage:** Required areas covered; GUI-layer scan lifecycle covered despite the exemption for that layer. `pytest tests/` — 1803 passed, 21 skipped, no failures. CI: `claude-review` ✓, `Tests + build verification` ✓.

PR #395, opened 2026-09-03
Fix Generate Enhancements crash on single-category runs (#389)

### LINKED ISSUE #389: Smart Citizen 2.3.1 crashes during Generate Enhancements — 0xC0000374 HEAP_CORRUPTION
Hi,

I encountered a reproducible crash in Smart Citizen v2.3.1 while generating enhancements. I was able to isolate the trigger and build/test a local fix.

Environment
Smart Citizen: v2.3.1
Windows x64
Python: 3.11.9
Star Citizen: LIVE
Problem 1 — native crash during Generate Enhancements

Running Generate Enhancements consistently caused Smart Citizen to terminate without a normal Python exception.

Windows exit code:

-1073740940

which corresponds to:

0xC0000374 / STATUS_HEAP_CORRUPTION

With Python faulthandler enabled, the process was terminating while EnhancementsGeneratorWorker was waiting on future.result() inside generate_enhancements_ini.py.

I tested the same enhancement generator independently from the PyQt GUI:

generate_enhancements_ini.main(..., max_workers=1)

using the same base.ini and the same DataForge cache.

The standalone run completed successfully with exit code 0.

This narrowed the trigger in my environment to the execution structure:

PyQt QThread -> ThreadPoolExecutor(max_workers=1) -> generator/XML processing

I also disabled the Qt progress callback as a test. The application still crashed with the same 0xC0000374, so the progress callback was not the cause.

Local fix

I modified the generator so that when:

max_workers == 1

it does not create an additional ThreadPoolExecutor thread and instead executes the submitted jobs synchronously inside the existing worker thread.

For max_workers > 1, the original ThreadPoolExecutor behavior is retained.

After this change:

Generate Enhancements completes successfully
the progress callback works normally
all enhancement generators complete
enhancement INI files are written successfully
Smart Citizen reloads the generated entries successfully
0xC0000374 no longer occurs
a newly built PyInstaller executable also works correctly

I repeated the full Generate Enhancements test with the progress callback restored and with the rebuilt .exe, and the crash did not return.

I cannot prove the exact low-level cause of the heap corruption, but in my environment the additional ThreadPoolExecutor when max_workers=1 is a reproducible trigger, and bypassing that extra thread resolves the crash.

Problem 2 — NoneType.category

I also encountered another loading exception:

AttributeError: 'NoneType' object has no attribute 'category'

in:

main_window.py -> update_category_combo()

at:

entry_categories = set(e.category for e in self.entries)

I added a defensive check:

entry_categories = set(e.category for e in self.entries if e is not None)

which prevents an invalid None entry from terminating the UI.

Both fixes are currently running successfully in my local v2.3.1 build.

I can provide the exact diff/patch if useful.

[crash_20260901_153810.log](https://github.com/user-attachments/files/31697533/crash_20260901_153810.log)

Stealrull:
Fixes #389.

## Summary
A user reported a reproducible native crash (Windows exit code -1073740940 / 0xC0000374 STATUS_HEAP_CORRUPTION) during Generate Enhancements, isolated to a `ThreadPoolExecutor(max_workers=1)` running inside a PyQt `QThread` worker — a thread nested under another thread doing lxml-based XML parsing. They also hit a secondary `AttributeError: 'NoneType' object has no attribute 'category'` in `update_category_combo()` once the entries list had already been corrupted.

- `scripts/generate_enhancements_ini.py`'s `main()` no longer constructs a `ThreadPoolExecutor` for exactly one lookup job or one output-generator job — a pool of one buys no parallelism anyway, so the single-job case now runs inline on the calling thread instead. The `max_workers > 1` path is unchanged. The output-dispatch logic (`if name == "components": ... elif ...`) was factored out of the executor loop into a shared step so both the single-job and parallel paths go through the same assignment code.
- `MainWindow._drop_none_entries()` generalizes the reporter's own defensive-check idea beyond the one site they found: filtering only at `update_category_combo()` isn't enough, since `_restore_pending_user_edits` reads `e.key` on every entry earlier in the same reload path and would crash there first whenever the pending-edits snapshot is non-empty, and the table model reads `self.entries` directly afterward regardless. The fix filters once, at the source — both places `self.entries` is ever assigned (the synchronous load path and the async `FileLoaderWorker` finished-slot) — so every downstream consumer sees a clean list. Logs a warning (tagged `#389`) only when something was actually dropped.
- Neither fix can undo the underlying native heap corruption if it still occurs; the `ThreadPoolExecutor` removal addresses the reporter's actual reproduction trigger, and the `None`-filter stops a corrupted entries list from also taking down the UI thread with a Python exception.

## Test plan
- [x] New `tests/test_single_job_skips_threadpool.py` — runs the real `main()` against a fabricated single-category job with `ThreadPoolExecutor` patched, proving it's never constructed, and separately confirms the single-job path still produces correct output.
- [x] New `tests/test_drop_none_entries.py` — empty/no-`None`/single/multiple/all-`None` inputs, order preservation, and the warning-only-when-something-dropped behavior.
- [x] `pytest tests/` — 1794 passed, 21 skipped, no failures.


Stealrull (commented):
## Review verdict: Approve

**What:** Fixes a reported native heap-corruption crash (0xC0000374) during Generate Enhancements on single-category runs by no longer constructing a `ThreadPoolExecutor` for exactly one job, and generalizes the reporter's own defensive-check idea into a single-source `None`-filter on the entries list (`MainWindow._drop_none_entries`), applied at both real `self.entries` assignment points rather than just the one site the reporter found.

**Issue acceptance check (#389):** both reported problems addressed. Problem 1 (the native crash) — the single-job branch now runs inline instead of through a pool, matching the reporter's own isolated trigger, backed by a test that proves the pool is genuinely never constructed. Problem 2 (`NoneType.category`) — the fix is stronger than the reporter's own single-site patch, since it filters before `_restore_pending_user_edits` reads `e.key` earlier in the same reload path, which their fix would have left crashing.

**File review (5 files, all Clean):**
- `generate_enhancements_ini.py` — traced both the `lookup_jobs` and `gen_jobs` single/parallel branches; the `gen_jobs` output-dispatch loop was correctly factored into one shared step covering both paths, not duplicated.
- `main_window.py` — confirmed `_drop_none_entries` is applied at both (and only) the two real places `self.entries` is ever assigned, in the correct order relative to `_restore_pending_user_edits`.
- Tests — `test_single_job_skips_threadpool.py` proves the branch is actually exercised (patches `ThreadPoolExecutor`, asserts non-construction) and that output is still correct; `test_drop_none_entries.py` covers the full behavioral surface of the filter function.
- `tests/CLAUDE.md` — found and fixed the same doc-drift gap already caught once in #385: two new test files had no catalogue entries. Added.

**CI note:** first run hit a transient, unrelated Windows access violation in `test_tab_scrollbar_placement.py` — confirmed passing locally (isolated and full-suite) and on all three of this session's other PRs; re-run passed cleanly. Not a defect in this PR.

**Test coverage:** `pytest tests/` — 1794 passed, 21 skipped, no failures. CI: `claude-review` ✓, `Tests + build verification` ✓ (after retry).

PR #396, opened 2026-09-03
Start Apply button green when nothing has changed since last apply (#387)

### LINKED ISSUE #387: Apply Enhancements Shouldn't Always Start Red
The Apply Enhancements button is always red when the app starts up, even if enhancements were applied in a previous session and no changes were made since then. There should be some logging of changes and saves that is checked to determine what state this button should be in.

Stealrull:
Fixes #387.

## Summary
- `_apply_dirty` was hardcoded `True` at construction with no way to clear it before the first real edit, so the Apply Enhancements button showed red on every launch even when the loaded state already matched what was on disk in the game's `global.ini` from a previous session.
- `MainWindow._build_apply_merged_dict(sources_dict, hierarchy)` extracts `apply_to_game()`'s merge-and-stamp logic (merge sources with user overrides, weave `[Owned]` tags, stamp journal entries, stamp the frontend version watermark) into a shared, side-effect-free builder, so the real write and the new startup check can never compute different content for the same loaded state.
- `MainWindow._entries_already_applied()` uses that builder — only on the very first load, gated on `_initial_load_done` so every later reload (regeneration, channel/language switch, import) still correctly forces the button red via the existing `_recompute_owned` → `_mark_apply_dirty` path — to compare the merged output against the currently-applied `global.ini`.
- The comparison itself (`_matches_applied_output`, new Qt-free module-level function) walks only the stock `base.ini`'s own key set, not every key `merged_dict` carries: `merge_ini_files` (the actual writer) never writes a key that file lacks regardless of what the merge produces for it, so comparing every merged key instead would report a false "dirty" on every launch for any profile with such a key, even immediately after a genuine apply.
- Falls back to the pre-existing conservative `True` (red) default on any doubt — no applied file yet, no stock `base.ini` loaded, or any error along the way. Getting this wrong in the green direction would hide a real pending change.

## Test plan
- [x] New `tests/test_apply_already_applied.py` — 8 cases covering untouched stock passthrough, an override already written vs. not yet written, a merged-only key outside the stock key set being correctly ignored either way, and a key missing from the applied file entirely.
- [x] `pytest tests/` — 1794 passed, 21 skipped, no failures.
- [x] Built a portable (`SmartCitizen-Portable-v2.4.0.zip`, combined with #383/#385/#386/#389) and confirmed manually: the Apply button now starts green on a second launch after applying once with no changes made since.


Stealrull (commented):
## Review verdict: Approve

**What:** The Apply Enhancements button previously started red on every launch unconditionally. Now, only on the first load, `MainWindow._entries_already_applied()` builds the same merged output a real apply would write (via a newly extracted, shared `_build_apply_merged_dict`) and compares it against what's currently applied on disk, starting the button green when they match.

**Issue acceptance check (#387):** both stated criteria satisfied — the button no longer starts red when nothing changed since the last apply, and the fix works by actually checking applied state rather than adding an unrelated change-log.

**File review (4 files, all Clean):**
- `main_window.py` — traced `_matches_applied_output` against `merge_ini_files`'s real write algorithm line-by-line: it correctly compares only the stock `base.ini`'s key set, since that writer never writes a key that file lacks regardless of what the merge produces for it — comparing every merged key instead would false-positive "dirty" on every launch for any profile with an enhancement-only key. Confirmed `_build_apply_merged_dict` is a pure, behavior-preserving extraction (frontend-stamp/journal-stamp tests still pass). Confirmed the call-ordering interaction with `_recompute_owned`'s unconditional dirty-mark is handled correctly via the `_initial_load_done` gate — every later reload still forces red as before. One disclosed trade-off, not a defect: this doubles the load+merge cost once at startup on the main thread, but that matches `apply_to_game()`'s own existing precedent for the identical pipeline (already synchronous there too).
- `src/gui/CLAUDE.md` — new design-decision entry accurately describes the mechanism, every symbol cited cross-checked against the code.
- `test_apply_already_applied.py` — 8 cases precisely scoped to the function's actual decision surface, including the core insight case (a merged-only key outside the stock set must not affect the verdict).
- `tests/CLAUDE.md` — catalogue entry added correctly from the start.

**Test coverage:** `pytest tests/` — 1794 passed, 21 skipped, no failures. Also manually confirmed via a real portable build (combined with #383/#385/#386/#389) before this PR was opened.

PR #398, opened 2026-09-03
Wire Simple mode's Apply button into the same dirty-state tracking (#397)

### LINKED ISSUE #397: Simple mode Apply Enhancements button is always green/active even if no changes made
Simple mode Apply enhancements should be red on first run like advanced mode then change to green and stay green if no changes are made even between sessions.

Currently it's always green and can be clicked even if no changes have been made.

Stealrull:
Fixes #397.

**Depends on #396** (#387) — this branch is stacked on top of it, since these fixes call methods #387 introduced (`_entries_already_applied()`, `_build_apply_merged_dict()`, `_matches_applied_output()`). Until #396 merges, this diff will show both PRs' changes combined; it'll narrow to just this PR's changes automatically once #396 lands on `release/2.4.0`.

## Summary
- `SimpleModeWidget.generate_apply_btn` was hardcoded green from construction, with no dirty-state tracking at all — always clickable regardless of whether there was actually anything to apply. Distinct from #387, which only fixed the Advanced-mode toolbar button; the two buttons are separate widgets and #387 never touched this one.
- `MainWindow._set_apply_btn_dirty` (the existing chokepoint #387 already routes every caller through) now also calls `simple_page.set_apply_dirty(dirty)`, so construction, edits, the startup check, and theme/language refresh all keep both buttons in sync for free. `SimpleModeWidget` gets its own `_refresh_apply_button()` chokepoint combining busy-state and dirty-state: busy always wins for enabled, dirty always decides color and tooltip.
- Caught and fixed a real construction-order bug along the way: the first propagation call happens inside `create_toolbar()`, which runs before `simple_page` is constructed a little further down in `setup_ui()` — despite `SimpleModeWidget`'s own class body appearing earlier in the file, that's definition order, not call order. Guarded with `hasattr`.

## Two more bugs found via manual portable testing (not caught by the test suite)
- **Reload re-dirtying itself right after a successful apply.** Simple mode's one-button flow reloads immediately after `apply_to_game()` succeeds (to refresh the hidden Advanced view), and that reload's `_recompute_owned()` call unconditionally re-marked the button dirty via `_mark_apply_dirty` — undoing what `apply_to_game()` had just correctly cleared, and re-arming the exit-time "unapplied changes" warning too. Fixed by extracting the #387 check into `_refresh_apply_dirty_after_reload()` and running it after *every* reload, not just the first at startup — also correctly clears `_session_has_unapplied_edit`. This is strictly more accurate generally: switching to a channel/language that's already fully applied now correctly shows green too, instead of assuming red.
- **Tooltip not reflecting dirty state.** After the color/enabled fix landed, the tooltip was still the same static string in both states. `_refresh_apply_button()` now sets it too, via a new `simple_mode.generate_apply_tip_disabled` key (all 8 languages) mirroring Advanced mode's own `toolbar.apply_disabled_tooltip` wording.

## Test plan
- [x] New/updated tests in `test_ui_mode.py` (dirty/busy/tooltip interaction) and `test_apply_already_applied.py` (`_refresh_apply_dirty_after_reload`, stub-driven, no Qt).
- [x] `pytest tests/` — 1800 passed, 21 skipped, no failures.
- [x] Manually tested via three successive portable builds (combined with #383/#385/#386/#387/#389) as each bug was found and fixed: initial wiring, then the reload-redirty bug, then the tooltip bug — confirmed fixed each time before moving on.


Stealrull (commented):
## Review verdict: Approve

**Note:** this branch is stacked on #396 (#387), since these fixes call methods #387 introduced. The diff currently shows both PRs' changes combined; it'll narrow to just this PR's own changes automatically once #396 merges. This review focuses on what's new here — #387's own portion was already reviewed and approved in #396.

**What:** Simple mode's Apply button was hardcoded green since construction, with zero dirty-state tracking. Wired into the same `_set_apply_btn_dirty` chokepoint #387 established, plus two real bugs found and fixed via manual portable testing during this same cycle: a reload-after-apply that silently re-dirtied the button (and re-armed the exit "unapplied changes" warning), and a tooltip that never updated with dirty state.

**Issue acceptance check (#397):** all three stated criteria satisfied — starts red on first run, turns and *stays* green when nothing's changed (the reload fix specifically ensures the "stays" half), and is no longer always-green-and-clickable.

**File review (14 files — #397-specific portions all Clean):**
- `simple_mode_widget.py` — `_refresh_apply_button()` correctly combines busy/dirty for color, tooltip, and enabled; confirmed backward-compatible with the old `set_busy`-only behavior for any caller that never calls `set_apply_dirty`.
- `main_window.py` — the `hasattr(self, "simple_page")` guard fixes a real construction-order bug (`create_toolbar()` runs before `simple_page` exists, despite `SimpleModeWidget`'s class body appearing earlier in the file — definition order isn't call order), caught by `test_branding_labels_layout.py` (the one test building a real `MainWindow`). `_refresh_apply_dirty_after_reload()` traces and fixes the actual reload-redirty bug chain correctly — running the #387 check on every reload instead of assuming "any reload means dirty" is a genuine accuracy improvement, not just a narrow patch.
- Tests — `TestRefreshApplyDirtyAfterReload` (stub-driven, no Qt) directly locks in the reload bug's root cause, including a guard against overcorrecting (a genuinely dirty reload must not clear a real session flag). `test_ui_mode.py`'s new tests cover the tooltip half specifically.
- 8-language `ui.json` — new `generate_apply_tip_disabled` key, programmatically confirmed used/present/non-empty in all 8, wording adapted from Advanced mode's own existing disabled-tooltip translations for consistency.

**Test coverage:** `pytest tests/` — 1800 passed, 21 skipped, no failures. Both real bugs were invisible to the automated suite and only found via actual manual testing in a built portable — exactly why that gate exists.

PR #400, opened 2026-09-03
Fix success dialog reporting 0 enhancements after a fresh Simple-mode apply

### LINKED ISSUE #399: Simple mode reports 0 enhancements if first generate enhancements window is skipped
When running simple mode if you skip this dialog:

Image

Then click Apply Enhancements you will get this popup saying it will generate them:

Image

and it does (confirmed in game on fresh install) however the report once finished says:

Image

Which is wrong, it should report the enhancements added.

Stealrull:
## Summary

Fixes #399. When the startup "Generate Enhancements?" dialog is skipped and enhancements are generated for the first time via Simple mode's own **Apply Enhancements** click, the success dialog reported 0 enhancements even though the game file itself was written correctly.

**Root cause**: Simple mode's one-button flow calls `apply_to_game()` before the reload that refreshes `self.entries` with newly-generated content (that reload runs afterward, to update the hidden Advanced view). The old count was built from `self.entries`, which on this path was still whatever loaded before generation ran — typically nothing tagged `"enhancements"` yet.

**Fix**: extracted `_count_enhancement_categories(sources_dict)`, which counts `sources_dict["enhancements"]`'s own keys via `StringEntry.extract_category` instead of `self.entries`. `sources_dict["enhancements"]` reflects exactly what `apply_to_game()` just merged (including any "Include discovered items" strip already applied to it in place), so it's correct regardless of `self.entries`' staleness.

Verified the new counting method is semantically equivalent to the old `entry.source_file == "enhancements"` filter for the normal case too: `src/parser/ini_parser.py`'s `load_source_files` builds `source_origin` (which becomes `entry.source_file`) explicitly excluding the `user` source from its hierarchy, so a user override on an enhancement-touched key never "steals" that key's `source_file` attribution away from `"enhancements"`.

This branch is independent of #387/#397 (based directly on `release/2.4.0`), not stacked.

## Test plan

- [x] `tests/test_enhancement_count_report.py` (new, 5 cases): empty/missing `"enhancements"` key, category breakdown accuracy, other sources ignored, post-strip dict counted correctly
- [x] Full suite green on top of `release/2.4.0`
- [x] Manual: fresh install, skip startup "Generate Enhancements?" prompt, click Apply Enhancements in Simple mode — success dialog now reports the real enhancement count (confirmed working by reporter)


Stealrull (commented):
## Final verdict — Approve (posted as comment; GitHub blocks self-approval)

**Issue acceptance check** (linked: #399):
- Success dialog now reports the real enhancement count instead of 0 when enhancements are generated for the first time via Simple mode's own click (skipping the startup prompt) — confirmed both by the new unit tests and by the reporter's manual portable test.
- Fix doesn't touch the actual game-file write path (which was already correct) — scoped precisely to the reporting bug.

**Issues found:** none blocking. One non-blocking note: `src/gui/CLAUDE.md` doesn't get a new design-decision entry the way #387 did, but this is a straightforward extraction/bug-fix rather than a new pattern, so the function's own docstring carries enough context.

**What was done well:** the root-cause explanation (staleness of `self.entries` vs freshness of `sources_dict`) is precise and testable; the fix reuses the existing `StringEntry.extract_category` categorizer instead of reinventing it; the semantic-equivalence verification against the old filter (via `ini_parser.py`'s `source_origin` exclusion of the `user` source) is solid due diligence.

Reviewed per `.claude/commands/review_pr.md`.

PR #402, opened 2026-09-04
Honor the configured cache folder in the DataForge cache migrator

Coerwyn:
migrate_dataforge_cache_to_local() computed its destination as a hardcoded %LOCALAPPDATA%\Smart Citizen\{channel}\cache\dataforge, while get_dataforge_cache_dir() honors the CACHE_DIR override. A user who pointed the Config tab's DataForge Cache Folder at another drive had the ~1.2 GB cache dragged back to LocalAppData on the next launch, silently undoing the setting on every start.

Resolve the destination through get_dataforge_cache_dir() so the override is respected, return early when source and destination are the same path, and drop the empty destination directory that getter creates so the move renames instead of nesting a dataforge/dataforge.

Honoring the override makes a cross-volume move the normal case, and shutil implements that as a non-atomic copytree + rmtree. Stage the copy as dataforge.migrating and rename it into place, so an interrupted copy cannot leave a stamped-but-partial cache that the next launch would trust (dataforge_cache_is_fresh only checks the stamp, raw/libs and one XML) before deleting the intact source for it. Guard the destination lookup too: it mkdirs, so an override pointing at absent removable or network storage would otherwise raise out of startup before any window exists.

The CACHE_DIR -> portable -> LOCALAPPDATA resolution order is unchanged; this only makes the migrator consistent with it. Adds the first regression coverage for this migrator, including the back-compat path where an install with no override still migrates to LocalAppData and the destructive paths that can reach an rmtree.

### DIFF
--- a/src/utils/settings.py
+++ b/src/utils/settings.py
@@ -2812,30 +2812,47 @@ class AppSettings:
 
     @staticmethod
     def migrate_dataforge_cache_to_local() -> None:
-        r"""One-shot move of the DataForge XML cache from Documents → AppData\Local.
+        r"""One-shot move of the DataForge XML cache out of the user-data tree.
 
         Pre-1.0 the DataForge cache lived inside get_cache_dir() (Documents\…),
         putting ~1.4 GB of extracted XMLs into the OneDrive sync tree. Moving it
-        to AppData\Local eliminates per-file OneDrive / Defender / Indexer hooks
-        during extraction and keeps large build-artefact files out of cloud sync.
+        out eliminates per-file OneDrive / Defender / Indexer hooks during
+        extraction and keeps large build-artefact files out of cloud sync.
 
-        Idempotent: no-ops when the old path is already absent. If the new
-        location already has a valid stamp the old directory is cleaned up and
-        the migration is considered complete.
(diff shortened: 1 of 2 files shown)


Osiris-DevWorks (changes requested):
Looks good, just a couple of changes requested:

**1. Don't copy when an override is set.** When the paths differ, queue the old tree with `AppSettings.set_pending_cache_cleanup(old_dir)` instead of moving it, and let the normal freshness check rebuild at the configured location. `_cleanup_pending_old_cache()` (`src/gui/main_window.py:4664`) drains it after a successful re-extract, with a window up and `robust_rmtree`.

Reason: this migrator runs at `src/main.py:115`, before `QApplication` exists at line 187. Honoring the override makes a ~28k-file cross-volume copy the common case there — minutes of silent stall with no window, and it retries every launch if it fails. `config_tab.py:665` already made this call: re-extract, don't move.

**2. Scope the staging work to the no-override `Documents → %LOCALAPPDATA%` path.** Keep it — that path is already cross-volume for anyone whose Documents is on another drive, so the partial-copy hazard you found is real there.

The equality early-return and the `OSError` guard on `get_dataforge_cache_dir()` are both keepers as-is.

---

